PR #146, opened 2026-04-07
Consolidate shortcut recording into shared capture engine

AllDaGearNoIdea:
## Summary

I got confused trying to set a shortcut for invoking the launcher, and also when assigning shortcuts for individual items.

Inspired by Raycast, this PR:

- Uses a consistent, reactive interface for recording shortcuts in both Settings and "Change Shortcut" for individual items. This is a shared capture engine (`useShortcutCapture.svelte.ts`)
- This recording replaces the "Save" button, instead autosaving upon valid input (more keyboard-friendly, important for Raycast-like apps)
- When in recording mode, other input is ignored, while more liberally ingesting keys to accommodate a wider range of users (modifier combos and more normal keys allowed)
- Intuitively guides the user through requirements (at least one modifier and only one normal key) with inline errors for invalid keys (e.g. Backspace, Enter)
- Respects and incorporates Apple's HIG for key order: https://support.apple.com/en-gb/guide/applestyleguide/apsgf9067ae8/web#apdbee268aa29e04

Some QOL fixes:
- Conflict detection now across the Launcher Toggle and Individual Items, showing inline what already occupies the attempted combination
- Standardises `Control` (vs `Ctrl`) to match Rust and Tauri

## Test plan

- [ ] Settings > Shortcuts: record a shortcut, verify auto-save, try invalid keys, try a shortcut already assigned to an item
- [ ] Main launcher > Cmd+K on a result > "Assign Shortcut": verify modal opens with auto-recording, auto-saves on valid combo, shows conflict if clashing with launcher shortcut, Escape closes
- [ ] Cmd+Tab during recording: verify recording cancels and launcher toggle is restored
- [ ] Verify Shift+/ saves as `/` not `?`
- [ ] `pnpm test` and `cargo test` pass


Xoshbin:
Good work on this. Reviewed and merged with the following additions on top:

**Fixes applied during review:**

- `canonicalize_shortcut` helper added to Rust to normalise modifier order (Control, Alt, Shift, Super) before storing in the HashMap. Without this, shortcuts persisted in the old Super-first order would silently fail to fire after upgrade.
- `parse_shortcut` Ctrl alias removed — canonicalisation now handles it upstream.
- `get_valid_shortcut_keys` Tauri command added so `VALID_KEYS` is sourced from Rust instead of a hardcoded TS list that could drift.
- CSS variable fallbacks removed and hardcoded pixel values replaced with design tokens.
- `.kbd` span replaced with the `KeyboardHint` component.
- Colocated tests added for `useShortcutCapture.svelte.ts`.

**Bug fixed post-review:**

Shift-only combinations (e.g. `Shift+/`) were accepted as valid shortcuts because Shift is a modifier. On a US keyboard this means `?`, `!`, `@`, `(` etc. could accidentally trigger a registered item shortcut from anywhere in the system. Fixed by requiring at least one of Cmd/Ctrl/Alt alongside Shift — the same rule Raycast enforces.

PR #147, opened 2026-04-07
Improve settings discoverability

AllDaGearNoIdea:
## Summary

Settings was only accessible via the action panel (Cmd+K), or from the menu bar. This makes it easier to find:

- **Cmd/Ctrl+,** keyboard shortcut opens the settings window (standard macOS convention)
- **"Settings" appears in search results** as a built-in command, so typing "settings" in the launcher finds it directly

## Test plan

- [ ] Press Cmd+, (or Ctrl+, on Linux) — settings window opens
- [ ] Type "settings" in the launcher — "Settings" appears as a search result with "Asyar" label on the right
- [ ] Select the "Settings" result — settings window opens and launcher hides
- [ ] `pnpm test` passes


Xoshbin:
Thanks for the contribution — the goal is solid. Applied a few small fixes before merging: corrected the manifest name/description/type fields, swapped the dynamic import for the existing static commands reference, and added test coverage for the new keyboard handler.

Xoshbin:
@AllDaGearNoIdea  Hey, I noticed your commits are showing up with the default GitHub icon. You might want to check your local git config (git config user.email) and make sure it matches an email verified on your GitHub account so you get the contributor credit!

AllDaGearNoIdea:
Thanks @Xoshbin I've updated by global settings and pushed a correction to the open PR

PR #158, opened 2026-04-09
feat: add compact launch view mode

AllDaGearNoIdea:
Add a 'Launch View' setting with Default and Compact options.

Raycast calls this Window Mode, in the General section of their settings.

Compact mode works thusly in Raycast, with the behaviour mirrored here:
when the app is invoked, it only shows the search and bottom bar, not the results list
the bottom-right buttons are changed to a 'show more' prompt with a down arrow key cap

From a fresh invocation:
if the user presses down, the launcher essentially turns into 'default view' until dismissed
if the user types something, the list is shown, if they delete this text, the list is hidden
interaction with list elements is blocked, unless the list is visible

To ensure consistent sizing I have added `set_launcher_height` and positioned the top of launcher as a percentage from the top of the screen (same as spotlight and raycast)

I've added the setting to Asyar Settings / Appearance / Launch View: Default or Compact. Note, I omitted a sub-header.

Added tests for compact idle keyboard behaviour.

## Test plan

- [ ] Toggle between Default and Compact in Settings > Appearance > Launch View
- [ ] Invoke launcher in Compact mode — only search bar and bottom bar visible
- [ ] Press down arrow — expands to full default view until dismissed
- [ ] Type a query — results list appears; delete all text — list hides again
- [ ] Verify Enter, Cmd+K, and click are blocked while list is hidden
- [ ] Verify Escape, Cmd+, still work in compact idle
- [ ] Switch back to Default mode — launcher behaves as before
- [ ] Run `npm test` — all 971 tests pass


Xoshbin:
Thanks for the great PR — compact mode is a really nice addition and the
implementation is clean. I pushed three small follow-ups on top of your commits:

- Added input validation to `set_launcher_height` (NaN / bounds guard)
- Refactored the bottom-bar buttons into a reusable `BottomBarButton`
  component (also cleaned up a pre-existing raw button in the same file)
- Replaced the `<style>` block in AppearanceTab with Tailwind's built-in scale

PR #201, opened 2026-04-16
Fix Compact Launcher: Eliminate unwanted interstitial frames

AllDaGearNoIdea:
Please feel free to rip this apart or suggest other ways of achieving these things. I've spent a week battling the system to get to this point. It works, but may not suit your style or preferences.

I use Raycast's compact view mode and it is instant. We merged in my PR ([#158](https://github.com/Xoshbin/asyar-launcher/pull/158)) of the feature last week.

But I noticed many frames which were perceptible and unwelcome, unlike Raycast's seemingly instant switching.

I've tested and built this for macOS, so the underlying issues and fixes may vary.

Firstly, the window expanded before Tauri could catch up:
expand too quick

Sometimes, it'd expand and show the new bottom bar without results:
new bottom bar in place of results

After the window expanded, it'd show the default list, before the filtered one:
unfiltered list

When collapsing, the window would resize before showing the "show more" bottom bar, leaving the top of the results visible:
collapse before bar change

All of the above would happen quickly, but made the whole thing feel clunky and certainly not smooth.

I've pushed all frame-synchronous work into Rust to keep the DOM still, ensuring AppKit and WebKit stay in sync.

One CATransaction per transition. Rust does the NSWindow resize, repositions the pinned webview + vibrancy layer, and toggles the "Show More" bar visibility inside a single main-thread block. Visually this happens in the same frame.

Changing the Window Mode option in settings is reflected the first time the user re-invokes the launcher, no visual re-organising.

Native "Show More" bar on macOS. An AppKit NSView sibling of the webview, so its visibility commits atomically with the window resize. I've tried to pull in theme colours from JS to match the webview palette.

Windows/Linux keep the existing Svelte overlay. I don't have access to these platforms to test there.

The webview is pinned at 560px and the Svelte tree uses a fixed 560px layout. NSWindow crops, so WebKit doesn't move things about.

settings.appearance.launchView is read synchronously during setup_app, so compact users land at 96px on the first show

Hiding after invocation works depending on action. If a query is active, hiding and re-showing brings it back. However, if the user presses "Show More" and escapes, a new invocation is in compact. This seemed most in-keeping with existing functionality while serving an experience one would expect after selecting "compact mode".

Worth mentioning...
  
JS seemed to be building the store path via string concat, which was writing to a different file than Rust opened on some Tauri versions. I switched this to path.join so both sides see the same store.

As a side-effect of pinning the webview on top of vibrancy, the contentView handles corner rounding, which correctly also rounds the bottom (ala Raycast). I imagine this was always intended?

Primarily, I've got this working solidly on macOS, while Win/Linux use the Svelte fallback.

Ideally these fixes would be shared across platforms, but I couldn't get the timing reliable from Tauri/Svelte alone. Hopefully you'll appreciate how smooth this feels...

**Testing**

On macOS, with compact mode chosen in settings:
- First launch lands at 96px without visual resizing
- Start typing, and at once: window expands, results appear, "show more" goes, "actions" appears at bottom
- Delete query, window collapses instantly
- Press "Show More" (down key), window expands instantly
- Press "Show More", then escape, then re-invoke, it should be compact and empty
- Type a query, then escape, then re-invoke, it should return to the same query view
- Change theme (dark/light or extension) and the Show More bar should follow

In either mode:
- Toggle launch view in settings, the next invocation should show the chosen mode instantly

In Default launch view:
- Unchanged

On Windows/Linux:
- Svelte fallback "Show More" bar still appears in compact mode prompt


Xoshbin:
Hey @AllDaGearNoIdea thanks for the PR.
It's strange you have these issues!
I have none of those issues you mentioned, otherwise I wouldn't merge the original PR.
Since the PR, I have been playing with the compact I didn't feel any issues and it was very responsive. I just tested out again comparing to Raycast it seems Asyar is more responsive than Raycast. I'm looking for someway to benchmark this.

I didn't test your PR yet, before testing your PR I just wanted to reach out to you that there maybe something wrong elsewhere. please ensure that all the three repos ([asyar](https://github.com/Xoshbin/asyar), [asyar-launcher](https://github.com/Xoshbin/asyar-launcher), and [asyar-sdk](https://github.com/Xoshbin/asyar-sdk)) are on the last main branch.
then in asyar-launcher repo run `rm -rf .svelte-kit node_modules/.vite` then in main project run `pnpm install` then `pnpm dev` to see if the issue maybe there.
I will test the current last main branch on all the platforms again (Macos, Windows, and Linux) and then test your PR against them.

Regarding showing the default list while searching, since there are a lot of changes and refactors happening, use the "reset search index" action command it's added for testing purposes regarding these issues while in development.

and from the screenshots you added I see awkward layout design issues, I will add current state of mine below to see the difference. please try to pull the latest main branch and test them until I come back to you with some more tests and results.
Screenshot 2026-04-16 at 10 43 42 AM
Screenshot 2026-04-16 at 10 44 06 AM

Xoshbin:
One more thing to try is to test the latest release builds that contains the compact view, either from the download page https://asyar.org/download or from the Github release https://github.com/Xoshbin/asyar-launcher/releases/tag/v0.1.1-14 that would tell us for certain it's either your development environment issue or no.

AllDaGearNoIdea:
It only happens briefly, but I'm quite sensitive to this stuff. I recorded the screen to check I wasn't imagining things.

I haven't timed it, but I think you're right that Asyar is more responsive. Raycast seems to take longer, but the view switches seamlessly. The PR I've put in still beats Raycast in speed, but eliminates unwanted interstitials.

Raycast slowed x16:
https://github.com/user-attachments/assets/ba91ba2b-ea34-453b-80b1-5d76cdf0d8d8

Asyar slowed x16:
https://github.com/user-attachments/assets/12c57122-c2be-4ebd-8858-8ed56b3c72bb

The above were recorded on an M2 Pro running macOS 26.4.1 with [v0.1.1-14](https://github.com/Xoshbin/asyar-launcher/releases/tag/v0.1.1-14). I've never developed on this machine, so it's an objective testing ground.

I see the same issues on my M3 dev machine. On that machine I've recorded by branch running.

PR #201 slowed x16:
https://github.com/user-attachments/assets/097deb23-53b0-4186-9619-7eb579e00bb5

If you'd prefer, I can create an Issue; I'm not wedded to the proposed solution(s), but hoping this research can help make Asyar both quick _and_ seamless.

Xoshbin:
That is really impressive! and I'm very happy to have a company with that sensitivity, because currently my focus is on the infrastructure and architecture of the launcher. I'm glad that your focus is on that part and your additions and pr's are very helpful for me and visually for end users.
when I saw the first two videos testing v0.1.1-14 my first impression was maybe the issue either in resultList.svelte or fetching the list from the rust side maybe slow.
then I saw your third video and saw the real difference which it's very impressive and very clean.
just give me some time to review the code then I will come back to let you know what to do.

Xoshbin:
Hey @AllDaGearNoIdea , I’ve made some tweaks to the PR to bring it in line with some recent architectural refactors I merged after your PR. I also noticed and fixed a few visual inconsistencies that popped up during the integration.

Before I merge, could you take a quick look and make sure everything still aligns with your original intent? I want to ensure I haven't accidentally broken any of the logic you implemented.

AllDaGearNoIdea:
@Xoshbin, that's still looking as intended. Great work, thanks for getting it in good shape.

Xoshbin:
Great, thank you for your contribution. amazing work.

Xoshbin:
That was interesting, the second I merged you commented.
I could not wait longer, because I was working on some new features and I did not want to waste my time with merge conflicts. thank you again.

AllDaGearNoIdea:
I hit "comment" pretty much exactly when you merged!

It's looking good, thank you.

I'm working on [149](https://github.com/Xoshbin/asyar-launcher/issues/149) now. I'll be able to do it, but I've noticed a couple of things we're missing (from Raycast) which would be good to include. I'll put in Issues for those, because they're "architectural" decisions you'd be better at implementing.

Xoshbin:
please open an issue I will work on it for sure.

Issue #149, opened 2026-04-08
Feature Request: User Defined App Directories


AllDaGearNoIdea:
I have .apps in idiosyncratic locations I'd like scanned. Raycast allows users to add additional paths for this purpose:
![Image](https://github.com/user-attachments/assets/30025052-606c-4b15-9f4f-98ed0060782c)

I'm happy to work on a PR, but wanted to run it past you first. Where would it make sense to add?

This feels like a core feature, rather than an extension. I see there's Search settings in the schema, but these aren't surfaced in the Settings UI (yet?).

What do you think?

Xoshbin:
hey @AllDaGearNoIdea 
That would be a good contribution, that was one of the features in the todo but not in priorities in the current stage but any contribution are very welcome.
to answer your questions
1- This is obviously a core feature, it's a core feature and It lives in Rust as a built-in, not as an extension. App search is wired directly into SearchState in Rust. it's not possible through extensions since the extensions are sandboxed and so far they don't have access to application informations.
And app discovery lives in Rust as a capability at asyar-launcher/src-tauri/src/commands/applications.rs
2- small correction: the AppSettings.search section in asyar-launcher/src/services/settings/types/AppSettingsType.ts currently has searchApplications, searchSystemPreferences, fuzzySearch, and enableExtensionSearch booleans. It does not yet have a custom-paths field — that part would be new. What you've noticed is that a couple of the existing booleans also aren't surfaced in the UI yet, which is a separate gap.

Since the application discovery where one of the very first features added in Asyar, it may need some refactors beside adding this new feature.
I'd be glad to see your pr, push your pr we will work on it together. you may add the feature then I may do the refactor on the same pr.
I'm not sure which platform you are using, but what i'm thinking about is since this is just about adding paths I think it may work across all the supported platforms. something to mention but not a big deal to worry about.
If you had more questions, please let me know.

Xoshbin:
Hey @AllDaGearNoIdea, just an update: I've refactored the settings page and extension architecture to align more with Raycast’s design. The foundation is now much better suited for this feature! As we discussed, the application discovery logic in Rust might still need a refactor, but it’s a much clearer path now. Let me know if you’re still interested in working on this, or if you'd like me to jump in and add it.

AllDaGearNoIdea:
@Xoshbin great, I'll take a look. Happy to work on it.

I've been working on a couple of extensions (Apple Shortcuts and KeePassXC), which are progressing nicely.

[...]

PR #221, opened 2026-04-17
feat(applications): custom scan directories

AllDaGearNoIdea:
Adds a section to Settings / Extension / Applications for adding custom app directories

Apps discovered outside the OS-default locations show the parent directory as the subtitle. When said dir is under $HOME, the prefix is replaced with `~/` - not sure what to do for Windows there...

The way paths are handled generally is intended to be platform-agnostic, but I've only tested on Mac.

I've changed the visual rendering of subtitle text to better match Raycast. I couldn't read it before.

Notes:
Raycast allows disabling built-in features, including Application surfacing completely. This seems to be blocked in code at the moment, but the toggles are visually shown. There are two toggles which seemingly do the same thing:
Screenshot 2026-04-17 at 20 08 35

I personally prefer your toggle interface, but Raycast only shows it in one place:
Screenshot 2026-04-17 at 20 10 02

Fine to have it in both places, but they should be linked.

Test via Settings / Extensions / Applications:
Add a directory present in defaults (e.g. `/Applications`), it should fail with an error 
Add a dir containing a `.app`, confirm it appears in a new search.
Search for an app in a custom location, should show with pretty subtitle.
Search for an app in `/Applications`, should have no subtitle.
Remove custom dir, confirm it doesn't show in a new search.
Verify other extension description subtitles render legibly.


Xoshbin:
This looks good! I’ll do a full review tomorrow.
Regarding the ui, please feel free to make design changes per your requirement or to bring it closer to Raycast's feel. because as I said my focus currently is on the infrastructure and api services, and I know the UI needs work that I won't be able to prioritize immediately. Any design improvements are very welcome.
If you ever plan to make ui changes, Please submit design changes in their own dedicated PR. Avoid combining them with feature additions or bug fixes so they are easier to review.
Regarding disabling built-in extensions, that is a feature needs to bee implemented.
thank you.

Xoshbin:
It always felt awkward to me to see the applications in the extensions tab in Raycast, so I took the opportunity and added a new tab dedicated to the applications. and now the enable button is working, when an app disabled it's not shown in the search result list anymore.
Your opinion matters, let me know if you think I took a correct path or no.
thank you again for your contribution.

AllDaGearNoIdea:
I’ve been thinking through the same thing, but haven’t settled on a solution.

I can see what they were going for, as the control of features is generally the same, but the approach of putting one long list in an Extensions tab _is_ confusing.

I’ve not taken a look at your updates yet, so may have the wrong impression, but I think the clearest delineation is between built-in features and extensions.

There’s two ways of approaching this and I’m not sure on the correct words to use, so take this more as an explanation than proposed terms.


1 Features tab, with dedicated (separate) spaces for Built-in tools, and Extensions.
-or-
1 Features tab, 1 Extensions tab.

What do you think?

Either way, having a dedicated space for actual Extensions (functionality that isn’t built-in) allows  folding more things in, like managing what’s installed and the options currently in Advanced.

Xoshbin:
I can clearly imagine what you mean, I'm more toward the "1 Features tab, with dedicated (separate) spaces for Built-in tools, and Extensions.". while still hard to decide yet. let's see where we end up.

Xoshbin:
Closes #149

PR #231, opened 2026-04-19
Improvement: Escape Key Options and Behaviour

AllDaGearNoIdea:
Requires [SDK PR #81](https://github.com/Xoshbin/asyar-sdk/pull/81)

Raycast has two escape key options which they describe thusly:
_"Preferred behavior for the Raycast window when the Escape key is pressed.

The default behavior is to clear the search → navigate back → close the window.

Change to "Close window and pop to root" to immediately close the window on Escape."_

Asyar currently has:
Close launcher (just hides window, leaves in current state)
Go back (from what I can see, does the same as close launcher)

I'm not sure what was intended, and can see some people wanting a "Hide Window" option that retains a user's place, so I've included that. But if that wasn't what you envisioned, I'd be comfortable only having essentially the two options from Raycast.

In this PR we have:
Reset Launcher (same as "Close window and pop to root")
Step Backwards (same as Raycast's default, and Asyar's new default)
Hide Window (as described above)

To fully emulate escape key behaviour, the launcher also restores and selects entered text when stepping backwards:
1. Invoke launcher
2. Type "show"
3. Press enter on 'Show Clipboard History'
4. Type "test"
5. Press CMD+K
6. Type "test"

Then press escape and look at the incremental behaviour, first it removes text, then hides (or navigates back from) the current window/panel; returning to the previous text-entry selects it, rinse and repeat.

That part needed a companion [SDK PR #81](https://github.com/Xoshbin/asyar-sdk/pull/81), which needs to be merged first. Once live, the lockfile needs a pnpm
install to pick up 1.15.11, then committed here.

pnpm vitest run, all 1300 tests pass.


Xoshbin:
Merged with some changes.
And I took the opportunity again to fix a long running bug in the action panel where the focus outside the action panel would not close the action panel.
Thank you for the PR again.

PR #237, opened 2026-04-19
Fix: Extension Display Upon Keyboard Shortcut

AllDaGearNoIdea:
There are a few subtle behaviours corrected here.

If the app is hidden and the user invokes an extension directly, it goes straight there, then

- if they press escape, it dismisses the launcher instantly (or hides if escape is set to hide).

- if they invoke another extension using a hotkey, it goes back to respecting the chosen escape behaviour.

This is how Raycast works. We have an additional "Hide" escape option in Asyar, and I believe a user who has selected that would expect escape to hide the window in all scenarios.

Similar to the launcher interstitial PR (#201), I've got rid of more here, such as the previous stale state showing when invoking an extension, or flashing the main launcher view when switching to an extension.

I also noticed that when already within an extension, then invoking another using its shortcut, the text-entry field wasn't automatically activated, so I've corrected that.

vitest: 1391 tests pass
cargo check and clippy: clean


Xoshbin:
Thanks for the PR — the flash-free reveal path, the replacement semantics, and the focus fix after a hotkey swap are all great additions. Merged with a few modifications and follow-up fixes.
One behavior change worth calling out: I dropped the hotkeyFromCold → hide-and-reset upgrade for the go-back escape setting.
I agree it matches Raycast, but I think since we have one more option than Raycast we should honor the setting literally. I myself prefer "step backward" option for my daily use, when I pick "step backward" it should pop the view and leave me at the launcher. previously it would go back to and hide the window which acted the same behavior as "reset launcher" option. so since we have more option users can select the "reset launcher" if they prefer this.
If this may not completely satisfy the intention you preferred in the first place, You may open an issue to discuss more options if any available.
Also fixed:
Iframe-focused extensions were swallowing Escape/Backspace/Cmd+K because iframe keystrokes don't propagate to the parent window — SDK now forwards host-reserved keys

PR #241, opened 2026-04-20
chore(gitignore): ignore src-tauri/built-in-features staging dir

### LINKED ISSUE #220: Gitignore src-tauri/built-in-features/ staging directory
This is a pretty simple one, but wanted to run it by you in case I'm overlooking something.

I noticed src-tauri/built-in-features/ is a staging dir regenerated every build copying manifest.json from each feature in src/built-in-features/

Since it's derived, it should be gitignored (only the old built-in-extensions path is ignored)

Any PR that touches a built-in feature's manifest commits the same content twice

I'd propose adding src-tauri/built-in-features/ to .gitignore, then git rm -r --cached to un-track existing copies.

Happy to do this myself, if deemed necessary.

AllDaGearNoIdea:
## Summary
- `src-tauri/built-in-features/` is regenerated every build from the per-feature `manifest.json` files under `src/built-in-features/`, so tracking it means every manifest change is committed twice.
- Replace the stale `src-tauri/built-in-extensions/` ignore entry with `src-tauri/built-in-features/`.
- `git rm -r --cached` the 11 existing copies so the index matches.

Closes #220.

## Test plan
- [ ] `git status` is clean after a fresh build (staging dir regenerates but stays untracked)
- [ ] `git check-ignore -v src-tauri/built-in-features/calculator/manifest.json` resolves to the new rule
- [ ] App builds and loads built-in features normally


Xoshbin:
Thank you
closes #220

Issue #219, opened 2026-04-17
Feature Request: Dynamic Data Sourcing


AllDaGearNoIdea:
I'm working on [#149](https://github.com/Xoshbin/asyar-launcher/issues/149) and noticed that if I make a change to one of the scanned directories, it isn't reflected in the app. This can only be rectified by restarting, or removing and then re-adding the directory.

Similarly, I'm working on an Apple Shortcuts extension, which needs to reflect the shortcuts actually available, not a stale list captured at launch.

This is especially important for an always running launcher app.

I've proposed this as a general-purpose thing, but maybe the app dir thing would be done in rust, and the extensions would be in the sdk?

Xoshbin:
We have the same issue when installing a new application in the os, Asyar is not aware of the new application until it's restarted.
Added it into the todo list.
Thank you

Xoshbin:
#242 fixes the application part of the issue.
I'm now moving to fix the other part, the extensions.
I will let you know when implemented.

Xoshbin:
Thanks for opening this issue. Catching this during beta allowed me to implement a necessary architectural overhaul #260 of the extension management system that would have been nearly impossible post-stable release. This change establishes a solid foundation for the ecosystem moving forward.

The underlying problem: the launcher's old single-iframe model evicted extension iframes after ~7 minutes of inactivity. Any extension subscribed to file changes (or any push event — applicationIndex, appEvents, systemEvents, fs:watch) would silently lose every event the moment the user dismissed the panel for too long. By the time the user re-opened the launcher, the extension had no idea anything had happened.

What changed: Tier 2 extensions now run as two iframes instead of one  — a persistent headless worker (keep_alive: None, never evicted on idle) and an on-demand view that's free to be Dormant. Push events route to the worker, which keeps callbacks alive across panel dismissals indefinitely.

Specific to your use case (fs:watch for Apple Shortcuts):

fsWatcher proxy #263 moved into the worker SDK bag (asyar-sdk/worker); asyar-sdk/view no longer exposes it. The watch handle and its callback must live in the worker.
Pattern: worker calls fs.watch(['~/Library/Shortcuts/']), on change re-pulls Apple's shortcuts list output, writes the fresh list into extension state. View subscribes via state.subscribe('shortcuts.list', ...) and re-renders on its own schedule — works whether the view was Mounted or Dormant when the change happened.
[...]

PR #262, opened 2026-04-26
style(launcher): Raycast-style visual & UX polish pass

AllDaGearNoIdea:
General UI / UX improvements to help with people switching from Raycast.

The key chips are now consistent across bottom bars and results list, consolidated into shortcutFormatter

Corner radii rounder and more consistent

Row sizing shows more content and less noisily, hotkey hints only show when selected, spacing when reaching top or bottom item, no wrap around

Results list no longer wraps around at top/bottom, a small spacing cue above/below the last item instead.

No more redundant "loading indicator", or double-presenting the selected item, but a space to show errors.


Xoshbin:
Hey, thanks for the PR — really nice polish overall. Just one thing left to sort:

Arrow-key wrap. Could you keep the old wrap behavior in launcherKeyboard.ts? ↓ on the last item should jump back to the first, ↑ on the first should jump to the last. That's how Spotlight/Alfred work and it's been the launcher's behavior since day one (And I personally prefer this behavior). Please revert that part and the related test changes. or we may add an option in the settings to let users decide which behavior to pick.

Heads up — I rebased your branch and added a small commit that swaps your gap: 11px / 13px and w-[23px] h-[23px] for new --space-2-5 / --space-5-5 / --space-7-5 tokens. Same exact pixels, no design change, just so the values live in the spacing scale. Take a look and let me know if anything looks off.

Everything else (font weights, the show_more_bar tweaks, the chip rim, the selection style) stays as you proposed — looks good and I love it even though I'm looking after an original ui design for Asyar, but for this stage of the app it's totally accepted and preferred.

Be aware that i merged main into this PR beside the changes I made, the merge contains a very big refactor that may cause your setup and extensions to break, sorry about that I had to merge main to see your changes and to guide you about the style tokens. I will add more details in #219 regarding the refactor.

Thanks!

AllDaGearNoIdea:
Thanks. Yeah, that’s fine about wrap. I reckon as an option makes most sense. I wasn’t sure if/what was intentional, but thanks for clarifying.

I’d like to see the same thing actually! I thought it’d be good to start with a solid base, and then diverge. Have you got any ideas / mockups? Raycast hasn’t particularly innovated since the AI pivot and I think there’s an opportunity to push things forward with this project.

No worries. It looks like you’ve been busy. I’ll work around you, appreciate it and thanks again for reviewing and improving!

Xoshbin:
To be honest, I haven’t prioritized a unique UI yet; I've been using Raycast’s design as a reliable baseline to ensure the core services and features work as expected while keeping the codebase clean during beta. Once the foundation is solid, I definitely plan to diverge and find Asyar's own visual identity.
Your changes are incredibly valuable for the current state of the app. They move the UI from a rougher 'alpha' feel toward a much more professional look, which I really like. I'll likely stick with this direction for a while until a more distinct creative vision for the UI takes shape, so definitely feel free to keep these types of tweaks coming.
Thanks for the great work.

AllDaGearNoIdea:
I've tested and it looks good. I've resolved a couple of conflicts. I also noticed the scrollbar wasn't respecting light/dark mode on macos and added that to the PR

Xoshbin:
Hey @AllDaGearNoIdea 
Sorry I had to revert back the arrow-key wrap, but I'd absolutely welcome a new PR from you to add it back as an optional setting.
Thank you for this nice PR, Now the ui feels more professional.

PR #269, opened 2026-04-28
fix(compact): eliminate flash on goBack, dismiss, and Settings

AllDaGearNoIdea:
When in compact mode, I noticed some routes showing a flash of the full-height launcher:
- launch, go straight into extension (no query typed), and goBack to launcher
- launch, perform "show more", hide window, show launcher
- launching an app (or no view command)
- opening other windows (i.e. settings, ai settings) via the launcher


Xoshbin:
Thank you, it's merged.

PR #279, opened 2026-05-04
fix(applications): index custom scan dirs reliably

AllDaGearNoIdea:
Apps in custom scan directories weren't showing in search.

The application service was running its first scan before settings finished loading, so the Rust indexer always saw empty `additionalScanPaths`.

The file-system watcher was keeping two copies of the extras list. Tauri updated one, and debounce read the other, so newly watched paths never made it into the index.

Rmoving/re-adding a dir only triggered a rescan on the next FS event.

Now apps in custom dirs should appear at startup, and changes should be reflected immediately.

Test:
- Add a custom app dir in Settings, confirm relevant apps show in the launcher.
- Quit and relaunch, confirm the same apps appear.
- Remove the custom dir, confirm the apps no longer appear.
- Re-add the dir, confirm the apps re-appear.
- Drop a new `.app` into the dir, confirm it appears without a relaunch.

### DIFF
--- a/src-tauri/src/application/index_watcher.rs
+++ b/src-tauri/src/application/index_watcher.rs
@@ -93,6 +93,26 @@ pub fn sync_result_to_event(result: &SyncResult) -> Option<IndexEvent> {
     })
 }
 
+/// Shared, mutable list of user-configured extra scan paths. One handle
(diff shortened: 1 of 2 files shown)


Xoshbin:
Thank you, I merged it with some little changes.

PR #283, opened 2026-05-04
feat(appearance): apply user preference (light/dark/system)

AllDaGearNoIdea:
The appearance options seemingly didn't have any effect.

Now this allows the user to set light/dark explicitly, or opt to follow the system's theme.

I collapsed nativeBarSync's listener into themeMode so it's the single source of truth, and nativeBarSync goes back to being a pure IPC bridge.

While looking at the css, I noticed Settings light-mode was transparent (I think unintentionally), so it's now opaque like dark mode.

Test:
- Change settings to Light / Dark / System, they should all behave as expected.
- When in System mode, keep the settings window open and change the system theme. You should see Asyar change accordingly.

### DIFF
--- a/src/resources/styles/style.css
+++ b/src/resources/styles/style.css
@@ -152,64 +152,49 @@ html[data-platform="macos"] .app-root {
   height: 480px;
 }
 
-/* ── Dark Mode ──────────────────────────────────────────── */
+/* ── Theme ──────────────────────────────────────────────── */
+
+/* Each palette is defined twice: under `prefers-color-scheme` (scoped with
+   `:not([data-theme])`) for OS-tracking, and under `[data-theme]` for the
+   user's forced choice (set by services/theme/themeMode.ts). */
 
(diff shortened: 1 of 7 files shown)


AllDaGearNoIdea:
I’m afk now, but I meant to post a screenshot of [Antinote](https://antinote.io), which has a nice interface for setting themes by associating them with light and/or dark mode.

Xoshbin:
Thank you for the pr, it was a long running issue I couldn't give time to fix it.
Regarding the Antinote, I had to download and install it to see it. it looks a little bit crowded, I'm not sure if I like it! I think it's kinda overwhelmed.

AllDaGearNoIdea:
Fair enough about the visuals, but the ability to choose themes for light and dark mode is a smart idea.

The app should generally have light / dark settings (as is now working) regardless of theme, but if the user chooses a theme, there should be a way for them to say they want that to be in effect when in light (and) or dark mode.

Issue #295, opened 2026-05-06
[Feature] Add API Key field for Custom OpenAI-Compatible Endpoints


bluemeda:
Currently, when selecting the Custom (OpenAI-Compatible) provider option, there is no field available to input an API Key.
While some local providers do not require authentication by default, many third-party OpenAI-compatible services and secured self-hosted instances (like LocalAI or vLLM with LiteLLM) require a Bearer token in the authorization header.

### Expected Behavior
When "Custom" is selected as the provider:

1. A new input field for API Key should appear.
2. The launcher should include this key in the headers as Authorization: Bearer <API_KEY> when making requests to the custom endpoint.
3. f left blank, the request should proceed without the authorization header (for local/unsecured use cases).

### Steps to Reproduce

1. Open Asyar Launcher settings/configuration.
2. Select AI Tab Menu
3. Select Custom (OpenAI-Compatible).
4. Observe that only the Base URL and Model ID is available, with no way to provide a secret key.

Xoshbin:
Thank you for reporting the bug.

Xoshbin:
hey @bluemeda thank you for reporting it out. it's been fixed.
please don't hesitate to report and open issues if you have troubles or have suggestions.

PR #303, opened 2026-05-07
feat: API key support for Custom (OpenAI-compatible) provider

Xoshbin:
Adds an optional API Key field to the Custom provider, sent as
Authorization: Bearer <key> when set. Also normalises the base URL
so users can paste with or without /v1, sends Anthropic's required
browser-access header, and allows http://localhost in the Tauri
fetch allowlist so local LLMs (Ollama, LocalAI, vLLM) work.


Xoshbin:
Closes #295

PR #318, opened 2026-05-12
Feat/run-tracker

Xoshbin:
MCP and Scripts manifests already referenced icon:server and
icon:terminal; the SDK now provides them, so those surfaces stop
rendering as blank spans. Dynamic agent commands default to
icon:sparkles to match the Agents extension identity, and scripts
without an @asyar.icon header fall back to icon:terminal.

A new manifestIcons contract test walks every built-in manifest.json
via import.meta.glob and asserts each icon:<name> resolves through
SDK hasIcon() — future drift breaks the test suite instead of
silently rendering a blank icon.


Xoshbin:
Hi @AllDaGearNoIdea, I’m planning to merge this PR now to keep the project history manageable, as it has become quite large.

I wanted to check in first to see if you had any other suggestions. Since this is on my internal branch, it’s probably easiest if I merge this now, and then you can send over any further improvements or refinements in a new PR once main is updated.

Really appreciate all the high-quality work you’ve been putting into this!

AllDaGearNoIdea:
@Xoshbin no worries. I'm going to be busy this weekend and next, so please don't let me hold you back.

I pulled it down for testing, added a directory, and got blocked by a truncated permissions view:
Screenshot 2026-05-12 at 23 55 41

AllDaGearNoIdea:
I had a bit of time to do some testing. I've not been able to make any changes, but thought you may appreciate my impressions seeing as I won't be around much. I hope this is helpful...
<br>

**Script Display in Launcher List**

The name of the script when shown in the launcher is just the filename (looks good).

When it is running, it's currently the path, but it should just be that same title.

Also, when it is running, if that item is a favorite, it shows both that filename version AND the running path version, so it's duplicated. The script item itself in the schema should just have an on/off state for the one item with a clean title.

This is the same item shown twice even though it's one process running:
Screenshot 2026-05-14 at 23 33 47
<br>

**Display of Running Scripts**
The main launcher is the list of scripts, pressing enter into one should go into THAT script, not a sort of "list of launcher items".

The top (launcher) area should allow for typing and essentially grep the output below to filter the display to just the lines in the query.

The main section should show like:
"Running...
/full/path/to/script

[output-from-script]

Bottom-left should be the icon, title of script, status display (e.g. blue dot, "running"), could also have the timer here too.

Cancel should be an option in CMD+K (ideally CTRL+C to mirror a terminal), rather than a visible button in the display.
<br>

**Auto Close**
Yeah, having used it, I think the user should opt-in to auto closing upon success.
<br>

**Compact Mode HUD**

I had thought about this since, and without wanting to mix metaphors, I wonder if its cleaner to just show things that are in flight, and things that need attention, regardless of source.

So if there is a script running and an agent running, does it make sense to show them as two different processes? I'm now leaning more towards just showing "2 active" and also dropping the icon. I think a few statuses that have a dot and text will look a lot cleaner.
<br>

**Tab to Ask AI**

I get wanting to surface this feature, but the current approach feels heavy handed / "look at me".

The important thing, is to alert a first-time user to this feature so you could:
rank "Ask AI" as the top favorite so it shows as the first item (remember, it will be in Default (not compact) on first launch, so will be very visible)
show a "Tab" thing on the item somewhere, different to where we show the user assigned shortcuts - you could make it only show when selected, but it will be selected on first launch, so should help with discovery of that feature without permanently changing the typing section!

Crude mockup:
Screenshot 2026-05-14 at 22 59 05

<br>

**Sidenote on Ask AI**

When I get time, I'll try setting this up to talk to an openai compatible agent, so I can test this feature.
<br>

**Launcher Bugs**
If a script is "hot" (so shown as a suggestion at the top of the launcher) it is not selectable in the list, the first command is auto-selected and pressing up goes to the bottom of the list instead of the visible script item above it.

Opening Show Clipboard History is super slow now.

If in compact mode, tabbing to "ask ai" shows some jumping around of the ui during the transition.

"Ask AI" looks different depending on how it's invoked:
if tab then Purple Ask AI and "Query..."
if search, enter, then "Search..."

It should be the same view regardless of invocation route. Personally I don't see the need for the Purple thing, and the placeholder is better as "Query..."

AllDaGearNoIdea:
P.S. if Blue = Running in the expanded launcher, then it's confusing to have Green = Active in the compact one when they are talking about the same item being in the same state.

Xoshbin:
Thank you. I appreciate your inputs, they are all pushing asyar toward a better launcher.
I fixed most of the issues in the last commits pushed, some of them with little changes of my own and some maybe i'm not with it like the ask ai, for now I will keep the ask ai which I think it's more convienient and easier to use for end users including myself. 
some others as said fixed with my taste, so you may send a pr or open a new conversation we may talk about and as always you are convincing me with your nice ideas and thoughts. but for now I merge it as it is.
regarding the compact mode. I agree I will try to fix it your way.
thank you.

PR #346, opened 2026-05-21
Script Management Tweaks

AllDaGearNoIdea:
I managed to get some time before I go away. This is not perfect, but gives an idea of how I think things should work. I just thought I'd get this over to you while it's fresh and in sync with main.

### Fixed
truncation miscategorising scripts
^ the after that fix, make sure dismissing still works

### Changes
launching a script doesn't hide the launcher
ctrl+c for cancel run - universal behaviour in terminals on mac, windows, and linux
ctrl+d for dismiss result - universal behaviour in terminals on mac, windows, and linux
note on the above, I can't think of an example of someone using ctrl+c to copy in Raycast, but perhaps you might be able to think of one?
ctrl+c and ctrl+d are reserved so can't be assigned to the hotkey or a shortcut
show inline time when running a script, blut dot doesn't blink
remove words "Running" and "Done" from main launcher
hint updated to show AI prompt
combined HUD statuses in compact mode, separator added when both shown, change order of items (dot, status, count)
status icon colors updated to match HUD, launcher, script view
running items (scripts and agents) now go into status sections
running items are sorted by newest first
in expanded view, the selected item maintains focus when other items re-order

### Didn't get around to
optional inline results (don't show automatically)
script history section - allow cancelling and dismissing
settings section for scripts (things like "dismiss done after x minutes")
improved script view
- there should be a Scripts History item which would show the view you've made access
- pressing enter on a script should show just that script

### Visible bugs:
switching to Ask AI using TAB is visibly janky from the compact launcher
scripts don't seem to be getting ranked when run regularly
if a script is renamed, it does show up, but as the full path

### Question
don't show keyboard shortcut unless highlighted - I keep fixing this, are you unintentionally restoring it?

### Coincidental work in the Raycast universe
[A Technical Deep Dive Into the New Raycast](https://www.raycast.com/blog/a-technical-deep-dive-into-the-new-raycast)
[Unsung - “This is a common tell in web apps, and we did a lot of work to eliminate it.”](https://unsung.aresluna.org/this-is-a-common-tell-in-web-apps-and-we-did-a-lot-of-work-to-eliminate-it/)


Xoshbin:
Hey @AllDaGearNoIdea, as always, thanks for your great contributions!
I'm currently sick and away from my computer right now. I should be able to get back online tomorrow or in the next two days to review this PR.
Regarding the "don't show keyboard shortcuts unless highlighted" issue—when I looked into it last time, I realized that change actually came from one of your previous contributions rather than being a bug that reappeared, but I hadn't found a good place to mention it to you yet.
On that note, could you please ensure that future PRs are specific to a single feature or bug fix? It's much easier to review and track changes when they aren't bundled together into a single PR, though I really appreciate how well you already document what each PR covers.
Thank you!

Xoshbin:
Hey @AllDaGearNoIdea, thanks again for the work here — there's some great stuff in this branch (the elapsed counter, the Ctrl+C/Ctrl+D shortcuts, the HUD cleanup). But I'm going to close this one without merging, for two reasons.

1. Scope. As I mentioned on your previous PR, please keep each PR focused on a single change. This one bundles at least eight independent features (section taxonomy redesign, HUD redesign, elapsed counter, run-row shortcuts, keepLauncherOpen, aiHintIntensity removal, selection pinning, native bar style fields). When they're combined like this it's very hard for me to review, and if one of them causes a regression I can't revert it without losing the others. Could you split these into separate PRs? I'm happy to merge them one at a time.

2. Two functional regressions in the current branch:

Active agents become invisible. The new isAttributed filter in searchResultMapper.ts suppresses the run row when a definition row exists, with the idea that the def row will carry the status. But statusForRow in itemStatusLogic.ts only handles cmd_scripts_dyn_* — it returns null for cmd_agents_dyn_*. So an actively-running agent has no run row (suppressed) and no status on its def row → invisible. Your new contract test at runRowVisibility.contract.test.ts:245 actually codifies this as expected behavior. This is the same regression PR #336 fixed.
Anonymous active runs lose their subtitle. buildRunMappedItem sets object_id: \run_${run.id}`butrunningStartedAtForRowcomparesr.id === item.object_id, so the lookup never matches for type === 'run'` rows. Combined with the subtitle being changed to empty (it used to say "Running"), anonymous runs render with no text at all.

Thanks again for the contributions — would love to see these come back as smaller PRs

Issue #449, opened 2026-07-06
Let extensions open non-web URL schemes (e.g. steam://) via a declared scheme allowlist


dose-dot-dev:
### Problem

Extensions can't open a registered OS protocol handler (custom URL scheme) without `shell:spawn`.
`tauri_plugin_opener.open_url` refuses non-web schemes. Observed at runtime when opening
`steam://run/<appid>` via `asyar:api:opener:open`:

```
[invokeSafe] plugin:opener|open_url: Not allowed to open url steam://run/3932890
```

Both extension-facing paths hit the same web-only allowlist (baked into Asyar's Tauri capabilities):

- `asyar:api:opener:open` → permission `shell:open-url` (`permissions.rs:137`)
- `asyar:api:browser:openUrl` → `browser/service.rs` → `open_url`

So launching anything by protocol URL (Steam/Epic games, `vscode://`, `obsidian://`, `zoommtg://`,
etc.) forces a `shell:spawn` fallback — console-window flash, "a script ran" run-tracker
notification, and a per-binary trust prompt.

### Proposed solution (preferred)

A **manifest-declared URL-scheme allowlist**, mirroring the `fs:watch` `permissionArgs` pattern:

```json
{
  "permissions": ["shell:open-url"],
  "permissionArgs": { "shell:open-url": ["steam"] }
}
```

The opener command checks the requested URL's scheme against the calling extension's declared list
before handing it to `open_url`; the declared schemes are shown in the permission prompt. This is
per-extension and auditable, versus globally whitelisting `steam://`.

### Alternative

A privileged `application:open(url|path)` capability that reuses Asyar's own launch path.
`open_application_path` (`commands/applications.rs`) already `open_path`s freely with host
privileges; a sibling that accepts a **scheme-gated** URL would let extensions launch protocol
handlers the same way Asyar launches app entries.

### Acceptance criteria

- [ ] An extension declaring `shell:open-url` with `permissionArgs["shell:open-url"] = ["steam"]`
      can open `steam://…`; a scheme not in its list is rejected.
- [ ] The allowed schemes are surfaced in the install/enable permission prompt.
- [ ] No `shell:spawn` is required to launch a declared-scheme URL.

### Impact

Combined with a `files:read` API (separate issue), this removes `shell:spawn` entirely from the
"index an app's files, then launch it by URL" extension pattern — no console windows, no run
notifications, no trust prompts.

---

[...]

Issue #448, opened 2026-07-06
Expose a permission-gated file content read to Tier-2 extensions (files:read)


dose-dot-dev:
### Problem

Tier-2 extensions have no way to read a file's **contents**.

- `fs:read` / `FileService.read()` is documented as "Future" (unimplemented).
- `IFilesService` (rewritten in #444) only searches the **index** — it returns paths + metadata via
  `files:search` / `files:status`, never file contents.

So any extension that needs to parse a local file (config, manifest, save data) must shell out via
`shell:spawn` (e.g. PowerShell `Get-Content`), which flashes a console window and raises a per-binary
trust prompt. Example use case: a Steam-games launcher reading
`steamapps/libraryfolders.vdf` and `appmanifest_*.acf` to enumerate installed games.

### The primitive already exists

#444 added `read_text_preview` in `asyar-launcher/src-tauri/src/commands/files.rs` — a bounded Rust
file read — with a rationale that matches this request exactly:

> `read_text_preview` exists because `@tauri-apps/plugin-fs`'s `readFile` is gated by the webview's
> fs capability scope … a Rust command instead, which isn't subject to that scope at all.

It just isn't reachable by extensions, for two reasons:

1. **Host-only.** It's a raw Tauri command (registered in `lib.rs`'s `invoke_handler`, invoked as
   `invoke('read_text_preview', …)` by built-in frontend code). It is **not** in the `asyar:api:*`
   permission gate — `permissions.rs` only added `files:search` / `files:status`. Sandboxed Tier-2
   iframes can't call raw Tauri commands.
2. **Scope.** `validate_path_allowed` (same file) restricts reads to **home / app-data / temp**. Many
   targets (e.g. `C:\Program Files (x86)\Steam`, other drives) are outside those roots and would be
   denied even if the command were exposed.

### Proposed solution

1. Add a `files:read` (or `fs:read`) permission and route `asyar:api:files:read` →
   `read_text_preview` through the permission gate so Tier-2 extensions can call it.
2. Make the read **scope declarable per extension** — the same pattern `fs:watch` already uses with
   `permissionArgs` globs — and show those globs in the install-time permission prompt. Concretely:
   relax `validate_path_allowed` from a fixed home/app-data/temp allowlist to "the fixed roots **or**
   the calling extension's declared, user-consented globs."

Keep the existing bounded-read safety (byte cap, lossy UTF-8, absolute-path + `..` normalization).

### Acceptance criteria

[...]

Issue #411, opened 2026-06-18
Feature: Support Indexing Windows Store / AppX GUI Apps


Runarry:
## Background

The current application index does not cover Microsoft Store / AppX apps. Users also cannot work around this by adding `%ProgramFiles%\WindowsApps` as a custom scan directory, because that directory is protected by Windows permissions and cannot be accessed directly by regular users.

## Detailed Description

Support indexing Windows Store / AppX GUI apps, so apps installed from Microsoft Store and built-in AppX apps can be discovered and launched like regular desktop applications.

Xoshbin:
Thanks for opening this issue and reporting this! I'll start working on this feature to support indexing Windows Store and AppX apps.

Xoshbin:
Hey @Runarry, thanks for reporting the issues! I've implemented the fixes and will be releasing a new version later today. Please test it out once it's live, and feel free to reopen this issue or open a new one if you run into any other problems.

Runarry:
@Xoshbin 
The current Windows Store / AppX app indexing still has two issues:
When the app starts and is activated for the first time, a command-line window briefly appears.
AppX app indexing is incomplete. Some installed apps that can be launched normally do not appear in search results, such as the Codex app.

Xoshbin:
Hi @Runarry, thanks for the detailed follow-up — both issues are fixed now:

1. **Console window flash** — the startup scan that enumerates Store/AppX apps runs a short PowerShell command. It was spawning a visible console window, which caused the brief flash on launch. It now runs hidden, so no more flashing.

2. **Missing AppX apps (e.g. Codex)** — some launchable Store apps were being skipped during indexing when their install location couldn't be resolved. That location is only used to find the app's icon, so it shouldn't have gated indexing at all. Now every launchable Store app is indexed (apps without a resolvable icon simply show no icon, but launch normally).

These will be in the next release. Once it's live, please give it a try — especially confirm Codex and any other previously-missing apps now show up.
Reopen or ping here if anything's still off. Thanks again for testing! 🙏

Runarry:
@Xoshbin 
I found a small issue and confirmed it locally:

PowerShell returns `Name`, `Aumid`, and `InstallLocation`, but the Rust-side `UwpApp` uses `name`, `aumid`, and `install_location`. This field name mismatch causes deserialization to fail, so AppX apps never make it into the index.

My solution is to add `PascalCase` mapping to the `UwpApp` data structure:

```rust
[...]

PR #450, opened 2026-07-06
Set CREATE_NO_WINDOW on extension shell spawns (Windows)

dose-dot-dev:
### Problem

Extension `shell:spawn` children are spawned without `creation_flags`, so every console-subsystem child (e.g. `powershell.exe`) briefly flashes a black console window. The `where` lookup in `shell::resolve_path` has the same issue.

The launcher's **own** PowerShell calls already suppress this — `application/service.rs` sets `CREATE_NO_WINDOW` on the UWP scan specifically to avoid the flash (#411) — but the extension-facing spawn paths in `shell/mod.rs` never got the same treatment.

Hit in practice by a Steam-games launcher extension that must use `shell:spawn` both to read Steam's library files and to open `steam://run/<appid>` (see #448 / #449 for why no non-shell path exists today): every background reindex and every game launch flashes a console window.

### Fix

Apply the same `CREATE_NO_WINDOW` flag (`0x0800_0000`) to both spawn sites in `shell/mod.rs`, `#[cfg(target_os = "windows")]`-gated, mirroring the existing pattern in `application/service.rs`:

- `spawn()` — set on the `std::process::Command` before conversion to tokio
- `resolve_path()` — set via tokio's inherent `creation_flags`

Stdio is piped in both sites, so suppressing the console does not affect output capture, and GUI-subsystem children are unaffected by the flag.

### Verification

- **Behavioral (Windows 11):** built the launcher from this branch and drove it with the Steam-games extension — PowerShell-based reindex and `steam://` game launch both run with **no console window flash**; output capture, exit codes, and the run tracker behave unchanged. Same actions on the unpatched launcher flash a console window each time.
- `cargo test` (Linux): 2472 passed; 1 pre-existing failure (`fs_watcher::registry_tests::rejects_when_global_total_paths_would_exceed_limit`) that reproduces identically on clean `main` under WSL's default `fs.inotify.max_user_instances=128` — unrelated to this change.
- `cargo clippy --all-targets -- -D warnings`: clean.

### DIFF
--- a/asyar-launcher/src-tauri/src/shell/mod.rs
+++ b/asyar-launcher/src-tauri/src/shell/mod.rs
@@ -277,6 +277,17 @@ pub fn spawn(
         .stdout(Stdio::piped())
         .stderr(Stdio::piped());
 
+    // CREATE_NO_WINDOW — keep a console-subsystem child (e.g. powershell.exe)
+    // from briefly flashing a black console window when an extension spawns it,
+    // matching the launcher's own PowerShell calls (application/service.rs, #411).
+    // Stdio is piped, so suppressing the console doesn't affect output capture.
+    #[cfg(target_os = "windows")]
+    {
+        use std::os::windows::process::CommandExt;
+        const CREATE_NO_WINDOW: u32 = 0x0800_0000;
+        child_process.creation_flags(CREATE_NO_WINDOW);
+    }
+
     let mut child = Command::from(child_process).spawn()?;
 
     let pid = child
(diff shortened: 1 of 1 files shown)


Xoshbin:
Thanks for this @geodose  — clean fix and it mirrors the CREATE_NO_WINDOW pattern we already use in application/service.rs for the same reason, so it's a good, consistent way to close this gap. Appreciate you tracking down both spawn sites (spawn() and the where lookup in resolve_path()) instead of just the obvious one. Merging now.

PR #457, opened 2026-07-08
feat(opener): declared-scheme allowlist for shell:open-url

### LINKED ISSUE #449: Let extensions open non-web URL schemes (e.g. steam://) via a declared scheme allowlist

dose-dot-dev:
Closes #449. Implements the corrected two-path plan from the issue thread: one declared-scheme allowlist, checked in Rust, gating both ways an extension can reach the opener — and `capabilities/default.json` stays untouched, web-only for host code.

### How it works

- **The allowlist**: web defaults (`http`, `https`, `mailto`, `tel` — the same set as `opener:allow-default-urls`) extended by the schemes in `permissionArgs["shell:open-url"]`, exact-matched lowercase, no globs. Schemeless/relative URLs are rejected outright. Bare `shell:open-url` keeps meaning the web schemes, so existing extensions using it for web URLs are unaffected.
- **`asyar:api:opener:open` now routes to a new `opener_open_url` Rust command** instead of the webview plugin command, so a declared scheme like `steam` launches its protocol handler without `shell:spawn` — permission check plus scheme re-check inside the command (same double layer as `fs:watch`), then the plugin's scope-free Rust-side `open_url`. Callers without an extension identity (privileged host context) get web defaults only — parity with the ACL this replaces.
- **`browser_open_url` passes the same gate for extension callers**, closing the hole confirmed in the thread: its `OsDefault` branch reaches the scope-free Rust `open_url` behind nothing more than `browser:tabs.write`. Host callers unchanged.
- **Manifest load-time validation**: `shell:open-url` args are *optional* (unlike `fs:watch`'s declared-iff-args rule); when present they must be valid lowercase RFC 3986 scheme names of ≥ 2 characters — a single letter is indistinguishable from a Windows drive-letter path (`C:\evil.exe` parses as scheme `c`) — and not `file`/`javascript`/`data`/`vbscript`, which launch content the OS treats as executable rather than a protocol handler. The scheme deny-list is a judgment call I'd flag for review.
- **Plumbing note**: getting the router-verified caller identity to the Rust commands means the `opener` and `browser` namespaces join `ALWAYS_INJECTS_CALLER_ID`, so every `BrowserService` method gains a leading `callerExtensionId` param — mechanical churn, but it's the pattern the registry drift-guard test enforces, and it keeps per-method special cases out of the router. Both `openUrl` wrappers use a raw `invoke` so scheme denials reject through to the calling extension instead of dissolving into a diagnostic.
- Consent surface (#455) renders the declared schemes via the existing `permissionArgs` plumbing; permission catalog and docs updated.

### Verification

- `cargo test` (2510+ passed; the one failure is the known `fs_watcher::registry_tests::rejects_when_global_total_paths_would_exceed_limit` under WSL's inotify limits, reproduces on clean `main`), `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt` clean, `pnpm -r test:run` green (new: scheme parsing/validation, allowlist checks incl. drive-letter and schemeless rejection, per-extension isolation, host-context parity, service/wrapper layers), `prettier --check` clean.
- Manual walkthrough on Windows with a real consumer — a dev-registered Steam-games extension declaring `permissionArgs["shell:open-url"] = ["steam"]`: the consent dialog renders the scheme chip; launching a game sends `steam://run/<appid>` through the gated opener — Steam starts the game with no console flash, no run-tracker entry, and no per-binary trust prompt; on a launcher without this PR the extension falls back to its `Start-Process` path.


Xoshbin:
Merged.
Thank you, your contributions adds a lot of values to Asyar.

PR #475, opened 2026-07-09
fix(cli): accept worker-only build output and fix Windows browser opening in publish

### LINKED ISSUE #474: asyar publish: worker-only extensions fail verifyBuildOutput; auth browser never opens on Windows
**Context:** First third-party `asyar publish` walkthrough (Steam extension, worker-only — `background.main` with no view), launcher v0.1.1-35 / SDK 4.1.0, Windows 11.

**Bug 1 — `verifyBuildOutput` rejects worker-only extensions**

```
✓ Validation passed
✔ Build complete
✗ Build output not found: expected dist/view.html (dual-entry) or dist/index.html (legacy web app) or dist/index.js (legacy library)
```

Two compounding defects in `asyar-sdk/cli/commands/`:

- `build.ts` `verifyBuildOutput`: `hasDualEntry = hasView && (!requiresWorker || hasWorker)` — a worker-only extension (`dist/worker.html` + `dist/worker.js`, no `view.html`, which is the layout `asyar-ext-builder` produces for background-only manifests) can never satisfy any accepted layout. Suggested: accept `requiresWorker && hasWorker` without `view.html` when the manifest declares no view-mode commands.
- `publish.ts` calls `verifyBuildOutput(cwd)` without the manifest argument (the `build` command passes it), so the worker logic can't even see `background.main`.

Locally patching both lets the publish flow proceed, so nothing else in the pipeline depends on `view.html` existing (and the launcher itself loads the flattened zip fine — `uri_schemes.rs` tries both `dist/<file>` and `<file>`).

**Bug 2 — auth browser windows never open on Windows**

Both `login()` and `getOrAuthorizeGitHub()` in `cli/lib/auth.ts` open the browser with:

```ts
exec(`${cmd} "${authUrl}"`);   // cmd === 'start' on win32
```

`cmd.exe`'s `start` treats the first **quoted** argument as the new window's *title*, so this opens an empty console window titled with the URL and no browser. Reproduced on both auth steps. Fix is the classic empty-title argument:

```ts
exec(`start "" "${authUrl}"`);
```

(The printed-URL fallback made both steps completable manually.)

Happy to PR both fixes.

dose-dot-dev:
## Summary

Fixes the three publish-pipeline CLI bugs from #474:

1. **`verifyBuildOutput` had no worker-only case.** `hasDualEntry = hasView && (!requiresWorker || hasWorker)` required `dist/view.html` unconditionally, so an extension with only `background.main` (no `mode: "view"` commands, hence no `view.html`) could never pass — `asyar build` and `asyar publish` both fail with "Build output not found". `view.html` is now required only when the manifest declares a view command; when no manifest is available the previous behavior is preserved. A missing `worker.html` also gets its specific error message in the worker-only case (previously that branch required `view.html` to be present).
2. **`publish.ts` called `verifyBuildOutput(cwd)` without the manifest argument** (the `build` command passes it), so even with (1) fixed, publish would still reject worker-only extensions.
3. **The auth browser never opens on Windows.** Both `login()` and `getOrAuthorizeGitHub()` ran `start "<url>"` — cmd's `start` treats the first quoted argument as a window *title*, so an empty console window opens and no browser. Both call sites now share an `openBrowser()` helper that uses `start "" "<url>"`.

## Testing

- New vitest coverage for `verifyBuildOutput` (view-only, worker-only, dual-entry, missing entries, no-manifest legacy call, legacy `index.html` layout) — 8 tests.
- Full SDK suite: 67 files / 588 tests pass; `tsc -p tsconfig.cli.json --noEmit` clean; Prettier clean.
- Verified end-to-end on Windows 11 with a worker-only extension: with these fixes, `asyar publish --dry-run` proceeds past the build check, and `start "" "<url>"` opens the default browser (the unquoted-title bug was confirmed live during the walkthrough in #474).

Closes #474

### DIFF
--- a/asyar-sdk/cli/commands/build.test.ts
+++ b/asyar-sdk/cli/commands/build.test.ts
@@ -1,0 +1,91 @@
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
+import { verifyBuildOutput } from './build';
+
+describe('verifyBuildOutput', () => {
+  let cwd: string;
+
+  beforeEach(() => {
+    cwd = fs.mkdtempSync(path.join(os.tmpdir(), 'asyar-build-test-'));
+    fs.mkdirSync(path.join(cwd, 'dist'));
+    vi.spyOn(console, 'log').mockImplementation(() => {});
+    vi.spyOn(process, 'exit').mockImplementation((code) => {
+      throw new Error(`process.exit(${code})`);
+    });
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+    fs.rmSync(cwd, { recursive: true, force: true });
+  });
+
+  function writeDist(...files: string[]) {
+    for (const file of files) {
+      fs.writeFileSync(path.join(cwd, 'dist', file), '<!doctype html>');
+    }
+  }
+
+  it('accepts a view-only extension with dist/view.html', () => {
+    writeDist('view.html');
+    const manifest = { commands: [{ mode: 'view' }] };
+    expect(() => verifyBuildOutput(cwd, manifest)).not.toThrow();
+  });
+
+  it('accepts a worker-only extension with dist/worker.html and no view.html', () => {
+    writeDist('worker.html');
+    const manifest = {
+      background: { main: 'dist/worker.js' },
+      commands: [{ mode: 'background' }],
+    };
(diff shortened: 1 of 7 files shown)


Xoshbin:
Thanks @dose-dot-dev for the sharp diagnosis and fix! Merged with small additions on top: asyar link now also passes the manifest (same worker-only bug as publish), and openBrowser() was hardened to execFile with an http(s)-only URL check, plus tests for both.

Issue #473, opened 2026-07-09
asyar publish: both OAuth steps request full `repo` scope (all private repos, read/write) — far more than publishing needs


dose-dot-dev:
**Context:** Walking through `asyar publish` for the first third-party store submission (the Steam extension, as invited in #460). I stopped at the authorization screens rather than approve them, because both grants are much broader than the pipeline needs.

**What happens today**

There are two authorizations in the publish flow, and both request the classic-OAuth `repo` scope:

1. **Store login** — `https://asyar.org/auth/github` redirects to GitHub with `scope=user:email,read:user,repo` (authorization-code flow, callback `https://asyar.org/auth/github/callback`). Because this is a server-side flow, a token with **read/write access to every public and private repo the user owns** is delivered to and held by the store server. GitHub's consent screen spells it out: code, issues, PRs, wikis, settings, webhooks, deploy keys — plus org resources. The `user:email` scope additionally grants the store read access to the user's **private email addresses**.
2. **CLI GitHub device flow** — `getOrAuthorizeGitHub()` in `asyar-sdk/cli/lib/auth.ts` requests `scope: 'repo read:user'`. This token at least stays on the developer's machine (keytar), but it's the same all-repos read/write grant.

**What the pipeline actually needs**

- The **store** needs identity only — the GitHub username, which the public `/user` endpoint returns with `read:user` (or even with no extra scope). Extension repos are public; ownership/release facts are verifiable without any repo scope at all. Private email addresses are likewise unnecessary: publishers stay reachable through their GitHub identity, and if the store wants a contact email it should ask for one explicitly at registration rather than read it from the account.
- The **CLI** needs to create a release and upload one asset on the extension's **public** repo (plus optionally create a public repo on first publish). The `public_repo` scope covers all of that — still coarser than ideal (classic OAuth has no per-repo granularity), but it never touches private repos.

**Suggested changes**

- Store OAuth app: drop `repo` and `user:email` from the requested scopes — `read:user` alone identifies the publisher.
- CLI device flow: request `public_repo read:user` instead of `repo read:user`.
- Nice-to-have: honor a user-supplied fine-grained PAT (e.g. `GITHUB_TOKEN` env var or a prompt) before starting the device flow, so cautious publishers can grant Contents read/write on a single repo and nothing else.
- Longer-term: a GitHub App with per-repo installation would give real least-privilege for both sides.

Happy to PR the CLI-side scope change if that helps.

[...]

PR #478, opened 2026-07-09
fix(cli): narrow GitHub device-flow scope to public_repo

dose-dot-dev:
## Summary

The CLI half of #473, per your review there: the device flow requested `repo read:user` — full `repo` grants read/write on every private repository the account can reach, far more than publishing needs. Publishing only ever creates a public repo, pushes to it, creates a release, and uploads an asset, so the scope is now `public_repo` — and `read:user` is dropped entirely since no CLI code path calls an endpoint that needs it.

Classic OAuth scopes are requested at runtime, so no OAuth-app reconfiguration is needed. Previously stored broad tokens can be swapped with `asyar publish --reset-auth`.

As you noted, the store-login scope is a server-side change on your end, so this addresses the CLI portion only.

## Testing

- `tsc -p tsconfig.cli.json --noEmit` clean; Prettier clean; SDK CLI suite passing.

Refs #473

### DIFF
--- a/asyar-sdk/cli/lib/auth.ts
+++ b/asyar-sdk/cli/lib/auth.ts
@@ -137,7 +137,12 @@ export async function getOrAuthorizeGitHub(): Promise<string> {
       },
       body: new URLSearchParams({
         client_id: GITHUB_CLI_CLIENT_ID,
-        scope: 'repo read:user',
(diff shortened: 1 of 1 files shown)


Xoshbin:
Thank you 🙏

PR #488, opened 2026-07-10
feat(cli): also honor GH_TOKEN, preferring it over GITHUB_TOKEN

dose-dot-dev:
## Summary

Follow-up to #479: the review there suggested also accepting `GH_TOKEN`, and I pushed that change to the PR branch — but a few hours after it had already been squash-merged, so it never landed. This PR carries that commit.

- `GH_TOKEN` is accepted alongside `GITHUB_TOKEN` and takes precedence, matching the gh CLI's documented ordering. The ordering is deliberate: in GitHub Actions `GITHUB_TOKEN` is the ambient workflow-injected token, while `GH_TOKEN` is only ever set on purpose — the explicit variable should win.
- Tests cover the fallback and precedence, and clear both variables in `beforeEach` so an ambient `GH_TOKEN` on a contributor's machine can't leak into the `GITHUB_TOKEN`-only cases.

## Testing

- Auth suite 10/10; `tsc -p tsconfig.cli.json --noEmit` clean; Prettier clean.

### DIFF
--- a/asyar-sdk/cli/lib/auth.test.ts
+++ b/asyar-sdk/cli/lib/auth.test.ts
@@ -67,24 +67,45 @@ describe('openBrowser', () => {
 });
 
 describe('getOrAuthorizeGitHub', () => {
-  const originalToken = process.env.GITHUB_TOKEN;
+  const originalGithubToken = process.env.GITHUB_TOKEN;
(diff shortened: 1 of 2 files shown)


Xoshbin:
Thank you 🙏.

Issue #498, opened 2026-07-13
cargo test on Windows: test harness fails to load — missing comctl32 v6 (TaskDialogIndirect) because the test binary carries no app manifest


dose-dot-dev:
# `cargo test` on Windows: test harness fails to load — missing `comctl32` v6 (`TaskDialogIndirect`) because the test binary carries no app manifest

## Summary

On Windows, `cargo test` in `asyar-launcher/src-tauri` compiles successfully but the test harness binary (`target\debug\deps\asyar_lib-<hash>.exe`) fails to *load*: the process exits with `0xc0000139` (`STATUS_ENTRYPOINT_NOT_FOUND`) before any test runs. The app itself is unaffected — `pnpm dev` builds and runs the launcher normally from the same tree.

Since CI runs the Rust suite on `ubuntu-latest` only, this never surfaces in CI; it means the Rust test suite currently cannot be run on Windows at all.

## Repro

Windows 11, stable Rust + MSVC Build Tools:

```powershell
git switch --detach origin/main   # reproduced at ac14ba49
cd asyar-launcher\src-tauri
cargo test
```

```
    Finished `test` profile [unoptimized + debuginfo] target(s) in 1m 57s
     Running unittests src\lib.rs (target\debug\deps\asyar_lib-<hash>.exe)
error: test failed, to rerun pass `--lib`

Caused by:
  process didn't exit successfully: `<repo>\asyar-launcher\src-tauri\target\debug\deps\asyar_lib-<hash>.exe` (exit code: 0xc0000139, STATUS_ENTRYPOINT_NOT_FOUND)
note: test exited abnormally; to see the full output pass --no-capture to the harness.
```

## What I ruled out

- **Not stale build artifacts** — reproduces identically after `cargo clean` (full rebuild).
- **Not a compile error** — the build finishes; the only warning is the pre-existing unused-`BOOL` `ShowWindow` warning in `window_management/windows.rs`.
- **Not the app binary** — `pnpm dev` (tauri dev) builds the same crate and the launcher starts and works normally.
- **Not a stray/shadowing DLL in `deps\`** — every `*.dll` next to the test exe is a compile-time proc-macro plugin (`serde_derive`, `tauri_macros`, `windows_implement`, …), not a runtime dependency, and each imports only standard system DLLs.
- Running the test exe directly exits with the same code (−1073741511) and no output; the usual "procedure entry point X could not be located in Y.dll" dialog is suppressed.

[...]

PR #504, opened 2026-07-13
test(files): make files:read/glob tests portable to Windows

dose-dot-dev:
Second slice of fixing the 38 Windows test failures that became visible once #499 made `cargo test` loadable on Windows (context in #498). This one is **test-only** — no product change (see "Why no product change" below).

## The failures

13 `commands::files` tests (the `files:read` / `files:glob` suites from #456/#460) built declared glob patterns from `std::fs::canonicalize()`'d temp roots. On Windows `canonicalize` returns a verbatim `\\?\C:\…` path, which breaks two ways:

- The `?` in the `\\?\` prefix is a glob metacharacter, so `files:glob`'s literal-prefix parse rejected every pattern (`must begin with an absolute literal prefix`).
- The pattern also carried native backslashes, which globset treats as escape characters, not separators — so `files:read` coverage never matched (`not covered by any declared pattern`).

A 14th failure was a mixed-separator assertion: a target path built by joining a forward-slash suffix onto a backslash root, compared for equality against the product's all-native output.

## The fix (test helpers only)

- Canonicalize temp roots with `dunce` instead of `std::fs::canonicalize` — no `\\?\` prefix — matching what the product already does to the requested path.
- Build declared-pattern strings through a small `slashes()` helper that forward-slashes the root. Requested paths and assertions stay native, so they still match the product's native output.
- Build the one exact-match target with sequential `.join()` calls so it uses the native separator.

## Why no product change

Declared `files:read` / `files:glob` patterns are forward-slash **by convention** — the product's own error text gives `'C:/Steam/appcache/**'` as the example, and the #460 review already evaluated and declined Windows-backslash pattern support (globset treats `\` as an escape). Normalizing backslashes in declared patterns product-side would reverse that decision, so the bug is purely in the test fixtures, which shouldn't have built patterns from `to_string_lossy()`.

## Verification

- Windows 11: `cargo test commands::files` fully green, including the 13 previously-failing tests.
- Linux (WSL): full `cargo test` green (known `fs_watcher` inotify env failure aside); `cargo clippy --all-targets -- -D warnings` and `cargo fmt` clean.
- Unix behavior is byte-identical: `dunce::canonicalize` == `std::fs::canonicalize` there, and `slashes()` has no backslashes to replace.

### DIFF
--- a/asyar-launcher/src-tauri/src/commands/files.rs
+++ b/asyar-launcher/src-tauri/src/commands/files.rs
@@ -595,6 +595,22 @@ mod tests {
     use super::*;
     use tempfile::TempDir;
 
+    /// Forward-slash string form of a path, for building *declared* glob
+    /// patterns. Extensions declare `files:read` globs with forward slashes
+    /// (the documented convention, e.g. `C:/Steam/**`), and globset treats
+    /// `\` as an escape — so a pattern built from a Windows path's native
+    /// backslashes would never match. Requested paths (not patterns) stay
+    /// native; the product normalizes those itself.
+    ///
+    /// Roots feeding patterns are canonicalized with `dunce`, not
+    /// `std::fs::canonicalize`, elsewhere in these tests: the latter yields a
+    /// verbatim `\\?\C:\…` prefix on Windows whose `?` is a glob
+    /// metacharacter, which breaks both pattern matching and the
+    /// literal-prefix parse `files:glob` relies on.
+    fn slashes(p: &std::path::Path) -> String {
+        p.to_string_lossy().replace('\\', "/")
+    }
+
     #[tokio::test]
     async fn test_write_and_read_roundtrip() {
         let app = tauri::test::mock_app();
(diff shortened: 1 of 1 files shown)


Xoshbin:
Thanks for the fix — merged. One non-blocking concern to keep in mind: the new slashes() helper converts Windows \ separators to /, so these tests no longer exercise native Windows-style glob patterns. Globset supports native backslashes on Windows; the actual failure came from the \\?\ prefix produced by std::fs::canonicalize, which dunce::canonicalize correctly fixes.  
This is fine for now, but please watch for native Windows-path regressions in future file/glob changes. We may later remove slashes() and build the test patterns with native PathBuf::join() calls.
And mostly after merging all the pr's that I have pushed commits to some of them, please if possible do a recheck to make sure nothing is broken on windows.

dose-dot-dev:
Good call - I'll definitely keep that in mind and run through some tests after all the merges. Thanks!

PR #512, opened 2026-07-14
ci: add a Windows Rust-test job (trial, per #511)

### LINKED ISSUE #511: CI: add a Windows job to run the keyboard-layout matrix and guard the Windows test suite
## Context

Until #499, `cargo test` could not even load on Windows, so the Rust suite has only ever run on `ubuntu-latest`. #499 fixed the loader failure and #503–#508 fixed the 38 Windows-specific test failures that then surfaced — `cargo test` now passes on Windows. But there is still no Windows CI, so those fixes (and Windows behavior generally) can regress silently.

## What a Windows CI job would do

1. **Guard the now-Windows-green suite.** The path-semantics, shell, and platform fixes in #503–#508 have no CI coverage on their target platform. A Windows runner keeps them from bit-rotting.

2. **Run the keyboard-layout matrix (follow-up to #506).** To avoid mutating a developer's OS input settings, the `windows_key_resolver` tests deliberately resolve only against *already-loaded* keyboard layouts and skip the rest — they never call `LoadKeyboardLayoutW`. On a typical machine only the US layout is loaded, so the AZERTY (`0x040C`), QWERTZ (`0x0407`), and US-International (`0x00020409`) assertions skip. A disposable CI runner can pre-install those layouts so the full matrix actually executes. This is the environment #506's design explicitly reserves for the complete matrix.

## Sketch

- A `windows-latest` job running `cargo test` (clippy/fmt are platform-agnostic and already covered).
- A pre-step that installs the layouts the keyboard tests look for — e.g. PowerShell `Set-WinUserLanguageList` adding `fr-FR`, `de-DE`, and the US-International keyboard for `en-US` — so `GetKeyboardLayoutList` reports them and `loaded_hkl_for_langid()` finds them. Mutating a throwaway runner's input config is fine; that is exactly the case #506 reserves for CI.

## Notes

- The keyboard tests are `#![cfg(target_os = "windows")]`, so today they compile only on Windows and run nowhere in CI.
- Independent of #506, which is complete on its own; this issue only adds the environment where its currently-skipped assertions run.

Refs #506, #499.

dose-dot-dev:
Implements the Windows CI runner from #511, which you green-lit for a trial. Since #499 + #503–#508 the suite passes on Windows, but CI is ubuntu-only — so those Windows fixes can regress silently and the keyboard-resolver matrix never runs.

## Build-time-conscious design (your main concern)

The Windows job runs **only `cargo test`** — no frontend/pnpm, no clippy/fmt/doc, since those are platform-agnostic and already covered by the ubuntu `test-and-lint` job. With `Swatinem/rust-cache`, the recurring PR cost is close to an incremental test-harness compile. If it's still too heavy, it's trivial to gate to nightly or `workflow_dispatch` — flagging that as your call after you see the trial numbers.

## Keyboard matrix loading is best-effort and self-reporting

The `windows_key_resolver` tests only resolve against already-loaded layouts (they never call `LoadKeyboardLayoutW`, so they can't touch a dev's input settings — that's #506). This runner is the disposable environment reserved for the full matrix: a `Set-WinUserLanguageList` step pre-installs French (AZERTY), German (QWERTZ), and US-International, and a diagnostic step (`cargo test windows_key_resolver -- --nocapture`) surfaces — through the tests' own `skipping: …` output — whether the session actually exposes them. If the runner's session doesn't pick them up, those cases just skip and the job still guards the rest of the suite, so the layout step can be tuned from the trial logs without blocking.

## What to watch during the trial

1. The Windows job's wall-clock vs the ubuntu job.
2. In the "Show whether the layout matrix loaded" step, whether the AZERTY/QWERTZ tests run or print `skipping`.

Note: I couldn't verify GitHub's Windows-runner behavior locally, so this PR's own run is the first real test — the layout-install step may need a follow-up tweak once the logs are in.

Closes #511.

### DIFF
--- a/.github/workflows/test-and-lint.yml
+++ b/.github/workflows/test-and-lint.yml
@@ -131,15 +131,21 @@ jobs:
           $list.Add('fr-FR')                             # AZERTY 0000040C
           $list.Add('de-DE')                             # QWERTZ 00000407
           Set-WinUserLanguageList -LanguageList $list -Force
+          # Cheap paper trail that the matrix's precondition is met; the
+          # resolver tests themselves run as part of `cargo test` below.
+          Get-WinUserLanguageList | Format-Table LanguageTag, InputMethodTips
 
-      # This is the first cargo invocation, so it bears the one-time compile;
-      # the keyboard diagnostic below reuses the artifacts and is near-instant.
+      # Runs the whole suite, including the windows_key_resolver matrix (the
+      # AZERTY/QWERTZ cases execute here because FR/DE were installed above).
+      # A single cargo invocation — a second one re-compiles the ~2-minute
+      # `asyar` crate from scratch, so the visible --nocapture diagnostic
+      # isn't worth doubling the job time.
       #
-      # One test is skipped here: `snippets::…routes_to_windows_resolver_for_
-      # shift_colon` asserts the *active* layout resolves Shift+`;` to `:` (US)
-      # via the foreground-window layout. This runner deliberately loads
-      # FR/DE for the resolver matrix, and per-user active-layout settings do
-      # not reliably apply to the test process in a headless session — so the
+      # One test is skipped: `snippets::…routes_to_windows_resolver_for_shift_
+      # colon` asserts the *active* layout resolves Shift+`;` to `:` (US) via
+      # the foreground-window layout. This runner deliberately loads FR/DE for
+      # the resolver matrix, and per-user active-layout settings do not
+      # reliably apply to the test process in a headless session — so the
       # active layout is non-deterministic here. That test still runs on
       # developer machines (US active); the resolver logic itself is covered
       # by the windows_key_resolver matrix, which reads GetKeyboardLayoutList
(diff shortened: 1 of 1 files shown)


dose-dot-dev:
Trial results are in — green, and tuned for build time.

**Windows job: ~5m10s**, on par with the ubuntu `test-and-lint` job (~5m). Getting there surfaced one real inefficiency: two separate `cargo test` invocations each recompile the ~2m14s `asyar` crate from scratch (they don't share the build), so my initial visible-diagnostic step had doubled the job to 7m16s. Dropped it — the resolver matrix runs inside the main `cargo test` anyway — and added `cache-on-failure: true` so cold re-downloads don't recur.

**Design notes:**

- **Lean by construction:** the job runs only `cargo test` — no frontend/pnpm-tests/clippy/fmt/doc, since those are platform-agnostic and already covered by `test-and-lint`. The unavoidable floor is the `asyar` crate compile (~2m), same on any platform.
- **Matrix runs:** `Set-WinUserLanguageList` installs French (AZERTY) and German (QWERTZ), and those resolver tests execute and pass on the runner — the #511 goal.
- **One CI-only skip:** `snippets::…routes_to_windows_resolver_for_shift_colon` reads the *active* foreground-window layout and expects US `:`. Loading FR/DE leaves a headless runner's active layout non-deterministic (per-user active-layout settings don't reliably apply to the test process), so it resolved `~`. I skip just that test in CI — it still runs on developer machines where US is active, and the resolver logic is covered by the matrix, which reads the loaded-layout list rather than the active one.

If ~5m per PR still feels heavy for the trial, the easy lever is gating this job to push-to-main + nightly rather than every PR (one-line change) — happy to do that; your call after you see how it runs.

Xoshbin:
Awesome work on this! Consolidating the test runs and avoiding the duplicate compilation overhead was a great call—bringing the job time down to ~5m makes it very reasonable to keep on every PR check.

Skipping the active-layout integration test on CI makes complete sense given the headless environment, especially since the matrix tests explicitly cover the same layout resolution logic anyway.

Merging this now. Thanks!

PR #556, opened 2026-07-18
feat(shell): seed binary trust from consent, fix settings enable/disable toggle

AllDaGearNoIdea:
I'm looking at my Apple Shortcuts extension and came across a few things.

The shell trust prompt renders in the launcher window, which is hidden while a worker spawns in the background, and the SDK's 10s IPC timeout kills before it can be answered. So a worker-only extension's binary had no way to ever get trusted.

**Shell trust from the consent dialog**
Extensions can now declare their binaries up front in the manifest, e.g. `"permissionArgs": { "shell:spawn": ["shortcuts"] }`. They show as chips in the existing install/enable consent dialog, and accepting seeds the shell trust store (bare names resolve via the same PATH lookup the spawn path uses). Revoking the extension's consent clears its trusted binaries. Undeclared binaries still fall back to the runtime prompt, so view-driven ones are unchanged.

**Settings enable/disable toggle**
While testing the above I found the extensions toggle in Settings doesn't actually work: it passed the display title to the id-keyed `set_extension_enabled`, and `invokeSafe` swallowed the NotFound, so the UI reported success while nothing changed. Even with the id fixed, the toggle runs in the Settings webview and the main window (which owns the extension host) never heard about it. Now the id is passed, failures surface, and Rust broadcasts `asyar:extension-enabled-changed` (mirroring `asyar:preferences-changed`) so the main window reloads. Disabling also purges the extension's manifest commands from the search index on the Rust side, so it leaves root search even if no reload runs!

Along the way: concurrent shell trust prompts now queue instead of fighting each other, an accepted consent remounts an already-running worker so `activate()` re-runs with the granted permissions, and `deactivate()` calls are time-boxed so one hung extension can't wedge future reloads.

Test:
- Install/enable an extension declaring `permissionArgs["shell:spawn"]`, confirm the binary appears in the consent dialog and a background spawn then runs with no runtime prompt.
- Revoke the extension's consent in Settings, confirm its binaries are gone from the Shell Trust Manager.
- Toggle an extension off in Settings, confirm its commands leave root search immediately; toggle back on, confirm they return and its worker remounts. No restart either way.
- cargo test and the launcher vitest suite green.


Xoshbin:
Thanks for this, @AllDaGearNoIdea — really solid work. 🙏

I put it through its paces for a good while: built a worker-driven harness in the sdk-playground extension (declaring permissionArgs["shell:spawn"] and spawning a binary from a background command) and walked every path in the PR:

Declared binary shows as a chip in the consent dialog; accepting seeds the trust store, and the background/worker spawn then runs with no runtime prompt — closing the exact gap you described (worker spawns while the window is hidden → the 10s IPC timeout killed the old prompt).
Revoking consent clears the extension's trusted binaries.
The enable/disable toggle now actually works — id fix + surfaced failure + the asyar:extension-enabled-changed broadcast + the manifest-command index purge all behave: commands leave/enter root search live, no restart, and the worker remounts on re-enable.
Correct on both fronts — the shell-trust-from-consent feature is well-designed (nicely symmetric with how fs:watch / shell:open-url args already work), and the toggle fix addresses a real bug. Merged. 

Thanks again!

PR #558, opened 2026-07-19
fix(shell): restore consent dialog and trust manager

Xoshbin:
Accepting the brief launcher flash because it only affects fallback consent.
Removing it would either delay every background command or require a larger architectural change. The dialog is now reliably usable, and declaring binaries avoids both the prompt and flash.

### UI/UX trade-offs

| UX area | Current behavior | Consent-aware architecture |
| --- | --- | --- |
| Normal background commands | Launcher disappears immediately | Launcher may linger briefly while the worker acknowledges execution |
| Undeclared binary consent | Launcher flashes away and returns | Launcher remains stable and shows the dialog directly |
| Perceived speed | Consistently snappy | Slightly slower or variable, especially during worker cold starts |
| Predictability | Always hides immediately | Hide timing depends on what the command does |
| Focus | Brief flash/focus interruption on the fallback path | Focus remains stable, but the launcher may hold focus longer |
| Failure cases | Consent dialog can cause a flash | A failed acknowledgement could make the launcher appear temporarily stuck |
| Accessibility | Sudden disappearance and reappearance can be disorienting | More continuous and easier for keyboard or assistive-technology users |

The main trade-off is:

> Better UX for the rare consent fallback, but slightly less immediate and less consistent hiding for every background command.

### DIFF
--- a/asyar-launcher/src/components/settings/ShellTrustManager.svelte
+++ b/asyar-launcher/src/components/settings/ShellTrustManager.svelte
@@ -1,8 +1,12 @@
 <script lang="ts">
   import SettingsSection from './SettingsSection.svelte';
   import { onMount } from 'svelte';
-  import { extensionManager } from '../../services/extension/extensionManager.svelte';
-  import { shellListTrusted, shellRevokeTrust, type TrustedBinary } from '../../lib/ipc/commands';
+  import {
+    shellListTrusted,
+    shellRevokeTrust,
+    discoverExtensions,
+    type TrustedBinary,
+  } from '../../lib/ipc/commands';
   import { feedbackService } from '../../services/feedback/feedbackService.svelte';
   import { logService } from '../../services/log/logService';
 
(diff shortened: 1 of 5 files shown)


Xoshbin:
Hey @AllDaGearNoIdea — following up on #556, I found two small issues:

- The Shell Trust Manager was always empty because Settings runs in a separate webview where `extensionManager.extensionRecords` is not populated.
- The fallback consent dialog for undeclared binaries could open after a background command had already hidden the launcher, leaving the dialog hidden and impossible to answer.

Both are fixed in this PR.

While fixing the consent dialog, I noticed another small UI glitch: the launcher briefly disappears and then immediately appears again when fallback consent is needed. This happens because background commands hide the launcher immediately, while the asynchronous consent request arrives slightly later.

I added the UI/UX trade-offs to the PR conversation because I know you have a sharp eye for these small glitches. Fully removing the flash would currently require a larger execution/visibility handshake, and that could make normal background commands linger or feel less consistent during everyday use.

For now, I prefer accepting the rare fallback flash instead of introducing a daily UI/UX cost for every background command. Declaring binaries avoids both the prompt and the flash anyway.

If we find a clean solution without those trade-offs, we’ll absolutely keep working on it. I just wanted to call this out so you’re aware of the behavior and why I chose to leave it this way for now.

AllDaGearNoIdea:
Ah, good spots. Thank you for letting me know. You understand the back-end better than me, appreciate it.

PR #584, opened 2026-07-27
feat(launcher): Nicer command argument behaviour and look

AllDaGearNoIdea:
While working on the Apple Shortcuts extension, I spent some time bumping against the command arguments UI/UX.

Commands with arguments now show ghost chip hints trailing the query, moving between fields with the keyboard works intuitively, and entered data persists until the text query changes. In general, this feels and looks a lot nicer.

### Notable changes
- **Argument persistence applies to drop-downs specifically.** `text`, `number` and `password` fields start empty every invocation and show their placeholder; a declared `default` is substituted at submit time rather than typed into the field. A hint you can type over beats a stale value you have to clear first. Docs updated accordingly.
- Enter is prevent only when a **required** field/argument hasn't received suitable (or any) data. Optional arguments being empty is fine, and doesn't slow the user down.
- `Run.silent` (`RunStartInput.silent`, `SpawnParams.silent`): is an explicit opt-in for internal plumbing to help prevent the user from receiving unwanted notifications. Defaults to `false`, so nothing changes for existing callers.

### QOL improvements that touch other parts
- `typeLabel` on dynamic command registrations: sets the right-side row label, so for example my Apple Shortcuts extension can label each item "Apple Shortcut".
- The ⌘K action panel and the argument drop-down list share one `.launcher-popup` surface, to keep things visually consistent and avoid duplication of code.

### Review feedback

- **#1, silent Tier 2 runs** — fixed with an explicit `silent` flag (see above)
- **#2, `mode: 'view'` flash** — the hide is now gated on `mode === 'background'`, incluode` case.
- **#3, hint cache** — invalidated on dynamic re-registration and on argument defaults being written
- **#4, submit hint / dropdown preview** — a refused Enter names the field in the bottom-left status area bar and via the same `enter()` seeding, so preview and results agree.
- **#5, untouched drop-down stashed.
- **#6, scope** — I've kept this in, as it's now opt-in

### Test
- Highlight a command with arguments, then Tab in and confirm nothing shifts visually.
- Arrow left and right to cycle them.
- Enter from a field or the main text entry both fire.
- Press enter when a required field is empty, the chip goes red and there's a bottom-left warning.
- Clicking around between chips, drop-downs, and other row items works as expected.
- With text entered, Escape back to the text entry / first field; type to change the query and see everything resets.

cargo test, clippy, fmt, and the launcher and sdk vitest suites all green.


Xoshbin:
Hey @AllDaGearNoIdea I just tested out the pr visually, I really loved the changes you made. anyways, I'm a little busy now I could not review the pr myself but gave it to claude and found some blockers that needs to be fixed. I'm happy to merge the pr after fixing those. thank you very much for the new look of the arguments.

## Review

Checked out locally and ran the full CI suite — everything green:

| Check | Result |
| --- | --- |
| `pnpm --filter asyar test` | 266 files, **2945 passed** |
| `cargo test` | **3127 passed**, 0 failed |
| `cargo clippy --all-targets` | clean |
| `cargo fmt --check` / `prettier --check` | clean |
| `svelte-check` | 127 errors — all pre-existing, none in the touched files |

The core feature is solid. Chips living inline in the search bar reads much better than the old command-chip row, the left/right arrow walk is well thought out, the Rust additions are all `#[serde(default)]` so nothing breaks on older payloads, and there's a new `ArgumentChipRow` component test covering the exact `bind:this`-in-a-loop bug that used to blank the launcher. Docs are updated thoroughly.

A few things I'd like addressed before merge.

---

### 1. :red_circle: The run-notification change silences *all* Tier 2 extension runs, not just plumbing

`asyar-launcher/src-tauri/src/commands/runs.rs:203` and `asyar-launcher/src/services/run/runService.svelte.ts:190` both gate on the same rule:

```
extensionId is set  AND  subjectId is missing  →  stay silent
```

The comment describes this as "internal plumbing (Tier 2 spawns)". But `subjectId` is **never** set on any Tier 2 path:

```
SDK  runService.start()   → RunServiceProxy.ts:15
                            sends { id, kind, label, cancellable } — no subjectId

Tier 2  shell.spawn()     → ExtensionIpcRouter.ts:235
                            appends only originRole — no subjectId
                            (shellService.svelte.ts:65-74 already documents this:
                             "Tier 2 callers ... never supply this")
```

So the condition reads as "plumbing" but actually means **"any run an extension started"**.

Concrete failure: an extension shells out to `yt-dlp` because the user pressed Enter on "Download Video". It fails. Today the user gets a `run_failed` toast and a failure row. After this PR they get **nothing** — no notification, no feedback, no row. The download just silently doesn't happen.

Two asks:

- Use an intentional discriminator rather than "absence of `subjectId`" — e.g. an explicit `silent` / `userFacing` flag on `runs:start`, or set `subjectId` on the paths that genuinely are user-facing.
- Split it out. It's a real behavior change in a different subsystem, and if the argument-entry feature ever needs reverting you shouldn't have to take the notification fix down with it.

### 2. :orange_circle: Tier 2 `mode: 'view'` commands will hide the launcher right after opening their view

`asyar-launcher/src/services/search/commandArguments.ts:36` hides and resets **unconditionally** after dispatch:

```
submit()
  └→ dispatchTier2Argument({ mode: active.mode ?? 'view' })
       ├→ dispatch(...)                          ← mounts the VIEW iframe
       └→ hideWindow() → resetLauncherState()    ← ...and immediately closes it
```

`mode` comes from the manifest, and `ExtensionLoader.ts:293` treats a **missing** `mode` as `view`. So a Tier 2 extension that declares `arguments` and omits `mode` gets a view that flashes and disappears.

No shipped extension hits this today — all five argument-declaring commands are `mode: 'background'` — so it's latent rather than broken. But it's a trap for whoever writes the next one. Gating the hide on `mode === 'background'` is a one-line fix.

### 3. :yellow_circle: `argHintArgsById` never invalidates

`asyar-launcher/src/routes/+page.svelte:297-310` memoizes each command's argument schema by object id, forever, with no eviction.

Dynamic commands are explicitly documented to keep a **stable id across renames and edits**, so a `replaceDynamicCommands` call that changes a command's argument schema will keep showing the old ghost chips for the rest of the session. The map also grows unbounded.

### 4. :yellow_circle: Two small UX gaps — your call, flagging rather than blocking

- **The submit hint is gone with no replacement.** The old `ArgumentChipRow` rendered `Enter · Run` / `Fill required fields`. Now, when a required field is empty, Enter silently does nothing and there's no explanation anywhere — and nothing was added to the bottom bar in its place.
- **The ghost dropdown chip previews the wrong option.** `+page.svelte:330` fills hint values from the stash only, so a dropdown previews as `data[0]`, while `enter()` (`commandArgumentsService.svelte.ts:225`) actually seeds the persisted/default option. The preview shows something different from what you get.

### 5. :yellow_circle: Escape with an untouched dropdown creates a stash and changes what Enter does

`commandArgumentsService.svelte.ts:264-269` copies every non-password field into the stash, including a dropdown auto-seeded from a persisted value the user never touched. So Tab → Escape with zero typing leaves a non-empty stash, which routes the next Enter through `runWithStash` instead of the normal Enter path.

Probably harmless in practice, but it fires when the user did nothing. Consider stashing only fields the user actually edited.

### 6. Scope

Three off-topic changes are bundled in. All three are authored by this PR — the concern is topic, not blame:

- the run-notification change (# 1 above)
- the nested-`pnpm-lock.yaml` rule in `.gitignore`
- the app-wide `::selection` style in `asyar-launcher/src/resources/styles/style.css:134-140`

The last two are harmless, so they're a nit.

Separately, the **argument persistence semantics changed**: text / number / password arguments no longer pre-fill from last value, only dropdowns persist. The docs are updated correctly and the reasoning is sound (a placeholder hint beats a stale value you have to clear first), but it's a user-visible behavior change that deserves a line in the PR description rather than being folded into "inline entry".

---

**Verdict: needs changes.** # 1 is the blocker — silent extension-run failures are a real regression against `main`. # 2 is a cheap guard worth adding while you're in there. Everything else is optional.

AllDaGearNoIdea:
@Xoshbin thanks, that prompted me to look into things a bit deeper. Still trying to focus on just this, but there were adjacent things. I'll have more PRs for other things related to getting my extension looking and running nicely, but will put those in separately when ready.

AllDaGearNoIdea:
@Xoshbin also please take notice of https://github.com/Xoshbin/asyar-coffee-extension/issues/1

Xoshbin:
Hey @AllDaGearNoIdea  Built it all and smoke tested it: `seed` with the three modes, the password rules, `default: ""` rejected, and the dynamic parity — `replaceDynamicCommands` now carries `requireAnyOf` and per-argument `seed`, so the gate is computed once for both paths like you asked.

Then I hit the thing you predicted, and I want your opinion on how I solved it.

## The problem

I tested `required` + `default` with a temporary manifest. It was bad:

- the chip showed the default, for example `world`
- Enter refused
- the bottom bar said "Value is missing in argument Who to greet"
- the chip turned red
- the only way to accept it was to type into the field

The value is right there on the screen, but the message says it is missing. Red means error, but nothing is wrong. And re-typing the same value to say "yes" is a strange thing to ask.

## What I did

Your words were "greyed until confirmed", so I built the confirm part:

```
Enter (first)   →  chips turn solid, bar says "Press Enter again to run with these values"
Enter (second)  →  runs
```

No typing, nothing red, and the pause is the point. Any edit cancels the prompt. A field that is genuinely **empty** still complains the old way — there is nothing to agree to there.

One rule I had to decide: **`requireAnyOf` is not satisfied by confirming.** Otherwise `caffeinate-for` would let you confirm `0 / 0 / 0` and fail again. `required` accepts a confirm, `requireAnyOf` needs a value you actually chose.

## And your grey question

You said if defaults are grey, we can not tell them apart from the field-name hints, which are grey too. You were right. I used a dotted underline:

| What you see | Meaning |
| --- | --- |
| grey, no underline | placeholder — a field name, sends nothing |
| grey, dotted underline | a seeded value — it **will** be sent, not agreed to yet |
| solid | the user's own value |

Both stay grey because neither came from the user. The underline says "this one is a real value". A value restored from a previous run renders solid, because that one *is* the user's.

I applied the same mark to the dropdown chip so both kinds of chip read alike.

## What I want from you

This is your UI, and I am not confident about the visual part.

1. Is the dotted underline the right signal, or do you have a better idea? I thought about italic and about a small dot before the value, but underline felt quietest.
2. Is two presses of Enter acceptable, or is it one press too many? The other option was to reject `required` + `default` in validation and say it is a contradiction.
3. The confirm message sits in the same slot as run failures, but in the accent colour with an info dot instead of red. Does that read as a question rather than an error to you?

Nothing uses `required` + `default` today, so there is no rush and it is easy to change.

AllDaGearNoIdea:
I think dashed is an intuitive way of communicating this?

I reckon these are our states:
Screenshot 2026-08-03 at 12 25 05

For the specific case of both a seeded and required field, those would initially show dashed.

If the user presses enter, the first required field would get selected (solid border). Either they type in a new value, or they press enter to confirm.

If there is another required field, that one then becomes the selection. Otherwise, the command fires.

This allows a user to "confirm" each required field relatively easily, while allowing them to easily change specific fields mid-flow.

So that sort of changes your questions, but:
1. Prefer dashed border
2. Don't need to block, the above solution to confirm each required field seems sensible
3. Keep the red error for the empty required fields case; with the above design, I don't think we need additional info text

AllDaGearNoIdea:
P.S. To handle tabbing, if a required field has been selected at any point during the current invocation, it counts as confirmed.

To make clear my intention, once a required field has been "confirmed" and is no longer selected, its border goes to none i.e. the state at the bottom of my list above.

AllDaGearNoIdea:
P.P.S. In the AI Chat design I'm working on, drafts (unsent prompts) have a dashed border too.

That'll mean Dashed = "not yet committed" feels like a consistent signal across the app.

AllDaGearNoIdea:
@Xoshbin I'll work on that today, I've decided to put the silence thing into its own PR, because opt-in to notifications makes more sense than showing everything all the time for internal plumbing.

Xoshbin:
Do you mean you'll do the dashed-border work, or the silent PR? Happy either way — just don't want us both building it.

AllDaGearNoIdea:
I meant I'll work on this PR by:
implementing the border
getting rid of the silencing - you were right, that deserves its own PR thread

Xoshbin:
Clear now, thanks. The border is yours, I will wait for your commits and not touch it.

AllDaGearNoIdea:
Okay I've done a final pass on this and implemented the enter = confirmation behaviour I described above.

I was wrong to suggest putting the requireAnyOf in a notification, it feels nicer in the app after trying it. I've incorporated that too, still requires at least one of the fields, but warns in-app with a dashed red chip.

The warning is simpler to read and allows for multiple different types of required fields in the small space.

It looks like this:
Screenshot 2026-08-03 at 18 22 16

As you can see, this is also working for scripts, which is how I tested it.

Here's a complicated example, with a required field, and 2 requireAnyOfs, with one of those currently selected:
Screenshot 2026-08-03 at 18 18 12

I know you said you preferred wrap for Asyar (my personal opinion differs), but specifically for the drop-down argument fields, it's really confusing to wrap-around, so I've made an exception there, I hope you can agree it makes sense in this specific context.

Please try all of this out and tell me what you think.

In general, this feels intuitive imho; complex things, communicated simply.

Again, the bot was correct in the original feedback about the silencing not fitting here, so I've removed it. Have you got an idea of how to implement that? Personally, I think specifically wanted notifications should be explicitly opted into, whereas background scripts running (to make things work etc.) should never bother the user and it feels wrong to require opting out in pretty much every extension. Errors (which hopefully will be rarely seen) should be handled differently, and we have that bottom-left space currently to surface those.

AllDaGearNoIdea:
My apologies @Xoshbin , I messed up the drop-down. It should now be working correctly.

Xoshbin:
This is really nice work, thank you. Selection as confirmation reads much better than what I had, and the in-app `requireAnyOf` warning is clearly the right call — you were right that a notification felt wrong there. I also like that you tested it through scripts and added the header support. Merging it.

Yes please, send the silencing as its own PR.

On your question — I think the answer is different for success and for failure, not different per extension:

**Success:** I agree with you completely. Background plumbing should be quiet by default. Asking every extension to opt out is the wrong way round, and the one that forgets is the one that spams the user.

**Failure:** this one has to stay a system notification. When a worker fires something in the background the launcher is usually not visible, so the bottom-left space is not on the screen for the user to see. `docs/explanation/run-tracking.md` already says this is the intent — a run "persists through view eviction and panel close" and "triggers a system notification on failure".

So maybe: success is opt-in to announce, failure always notifies. You get the quiet default you want, and a broken run can never disappear silently.

One small thing for the future, and it is honestly more about me than about you. I am slow at reviewing large PRs. This one grew to arguments, plus seeds, plus the silencing, plus a popup style refactor, and it took me a long time to get through it — I nearly missed the notification problem entirely. If a PR stays on one feature or one fix, I can review it properly and merge it much faster, which is better for both of us.

Thanks again, this turned out really well.

PR #593, opened 2026-08-07
fix: approve esbuild & keytar build scripts in both pnpm workspace configs

AllDaGearNoIdea:
Fresh checkouts fail `tauri dev` with `ERR_PNPM_IGNORED_BUILDS: esbuild, keytar` when pnpm 11 is on PATH.

My [#501](https://github.com/Xoshbin/asyar/pull/501) allowlist only used v11-only syntax, and `tauri dev` actually resolves the nested `asyar-launcher/pnpm-workspace.yaml` too, which had no allowlist at all.

This includes syntax for both pnpm 10+11, and is applied across workspaces.

Verified: a pnpm 11.20 install from `asyar-launcher` now runs both build scripts and completes, and `pnpm dev` through the 11.x shim gets all the way to vite. The pinned 10.26 flow is unchanged.

### DIFF
--- a/asyar-launcher/pnpm-workspace.yaml
+++ b/asyar-launcher/pnpm-workspace.yaml
@@ -2,3 +2,11 @@ packages:
   - '.'
   - 'src/built-in-features/*'
   - '../asyar-sdk'
+
+allowBuilds:
+  esbuild: true
+  keytar: true
+
+onlyBuiltDependencies:
+  - 'esbuild'
+  - 'keytar'
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -8,6 +8,10 @@ allowBuilds:
   esbuild: true
   keytar: true
 
+onlyBuiltDependencies:
+  - 'esbuild'
+  - 'keytar'
+
 supportedArchitectures:
   cpu:
     - 'arm64'


Xoshbin:
Hey @AllDaGearNoIdea Thanks for tracking this down — the nested-workspace diagnosis is right, asyar-launcher/pnpm-workspace.yaml really did have no build-script allowlist at all, so root's allowBuilds was invisible to anything run from inside that directory.

Looking at this again reminded us: deleting that nested pnpm-workspace.yaml was already the plan for consolidating onto the root workspace as the single source of truth — it just never got done. A few things point that way:

CI/release always run pnpm install from the repo root already (release-launcher.yml even overrides its job's default cwd back to . for the install step).
Only the root pnpm-lock.yaml is tracked — our .gitignore already has a comment calling nested lockfiles "stray artifacts of running install inside a package directory."
The nested file's src/built-in-features/* and ../asyar-sdk entries don't resolve to anything root doesn't already cover (root already lists asyar-launcher and asyar-sdk, and forces asyar-sdk: workspace:* via overrides).
If we just delete asyar-launcher/pnpm-workspace.yaml, pnpm install/tauri dev run from inside asyar-launcher/ should walk up and resolve the root workspace instead — picking up root's allowBuilds for free, no duplicate config needed.

Could you try that on your branch (pnpm 11 on PATH, fresh asyar-launcher checkout) and see if it resolves ERR_PNPM_IGNORED_BUILDS the same way? Want to make sure we're not missing a second issue your fix is also covering before we go that route instead.

Xoshbin:
Thanks — the build-script diagnosis was useful. We rechecked this against the pnpm 10.26 setup and verified that removing asyar-launcher/pnpm-workspace.yaml makes commands from that directory resolve the root workspace, where allowBuilds already permits both esbuild and keytar.
Since #593 duplicates workspace policy in two places and adds the deprecated onlyBuiltDependencies setting, we’re closing it in favor of the single-root-workspace fix.

PR #594, opened 2026-08-07
feat(launcher): render the Show More bar in the DOM on all platforms

AllDaGearNoIdea:
I've been doing some UI work and was struggling to port designs consistently acros to the macOS Show More bar.

Out of interest, I tasked Claude Fable with investigated methods to eliminate the interstitial flashes I'd struggled with in WebKit, but without the native macOS bar approach.

After a long while, it found a way and came back with this conclusion:
_The webview is pinned at 480pt and cropped by the window, so WebKit paints new states before the window moves, and every visible compact/expanded resize is now confirmed from the same rendering update that paints the swap. On macOS the resize is then committed inside the CATransaction that applies that paint's layer-tree commit, so window geometry and new pixels ship in one render-server commit. The mechanism (and why the obvious presentation callbacks can't do this) is documented above `SentinelPending` in `platform/macos/window.rs`. If the technique ever stops working, a 250ms watchdog degrades to the old one-frame behavior, never worse._

I've been testing and pixel peeping this and haven't managed to force or catch a stray frame. This PR removes almost 1500 lines for the native bar, including the sync bridges and push commands. With the Svelte bar the single source of truth across platforms, it should make work on the compact view easier. I think it's most important that things look correct and native, so wouldn't want a "good enough" fix. Having prodded this, I'd personally be happy with how it looks and feels.

On Windows+Linux, the bar is unchanged, but resizes are now requested at effect time, instead of after a double rAF. In theory, this should be slightly quicker than it was before, however there's no atomic-commit machinery on those platforms (from what I could see) to make it _as_ quick as macOS. I'd appreciate if peoeple could try this out on the other platforms to check this hasn't downgraded the experience in any way!

Tests: 3157 cargo, 3087 vitest, no new svelte-check errors.


Xoshbin:
Thanks for this one 👑

You are really the king of UI/UX here. These flashy one-frame glitches my eyes never see them, but you always do. Same as back in #201, when you made the bar native to kill those interstitials.

One funny thing I saw when I looked at the history: you built the native bar in #201, then the pre-commit gate in #269, then the paint confirm in #451 — and now you use all of those pieces to delete your own native bar 😄 So you spent 4 months building the tools to remove your own code. Respect.

I tested on Windows and Linux and found no problem. But only with my own eyes, no tools, no slow motion video. So hopefully one day we get a user on those platforms with eyes as sharp as yours 🙏

Honestly I don't fully understand every detail of the intent here, but I trust your intentions. Merging it.

PR #606, opened 2026-08-11
perf(timers): skip idle 1Hz SQLite poll

AllDaGearNoIdea:
powermetrics on a parked launcher showed the main process waking ~17 times/s at idle. One steady contributor was that the extension-timer scheduler queries SQLite every second even when nothing has ever scheduled a timer.

The registry now keeps a cheap `maybe_nonempty` flag, armed whenever a timer is inserted. The scheduler skips the poll entirely while it's clear, and the poll query itself reports (via an EXISTS over unfired rows, same single query) whether it should keep polling. A future scheduled timer will keep the loop alive / not get stranded. Firing behaviour is unchanged; this just removes un-needed wake ups.

Together with my next PR (perf(watchers)), idle main-process wake ups drop from 17.1/s to 11.5/s and CPU ms/s about 19% in 60s powermetrics windows on my machine. Summon latency is unchanged.

Test:
- Timer suite green (43 tests), including new coverage for the arm/skip path and the future-timer-keeps-polling case
- Schedule an extension timer → fires on time; remove all timers → poll stops
- Before/after powermetrics numbers above

### DIFF
--- a/asyar-launcher/src-tauri/src/timers/mod.rs
+++ b/asyar-launcher/src-tauri/src/timers/mod.rs
@@ -16,6 +16,10 @@ use crate::error::AppError;
 use crate::storage::DataStore;
 use rusqlite::params;
 use serde::{Deserialize, Serialize};
+use std::sync::{
+    atomic::{AtomicBool, Ordering},
+    Arc,
+};
 
 /// Tauri event name used when a timer fires. The TS-side bridge
 /// (`timerBridge.svelte.ts`) listens for this and dispatches the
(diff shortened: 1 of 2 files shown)


Xoshbin:
Thanks for the timer optimization. While validating the idle path, I found one edge case: once the final timer fires, the hourly cleanup of old fired rows could be skipped indefinitely. I added a focused regression test and patched it directly, so no action is needed from you. The 1 Hz SQLite polling remains disabled while there are no unfired timers. Please let me know if you disagree with this change or see a better approach.
I have been testing the PRs all day and added #612 to make it easy to benchmark the development build rather than the installed production app. I also found that cold start has increased by about one second sometime after the last release. After these PRs are merged, I will investigate where that regression was introduced.

PR #608, opened 2026-08-11
docs: idle-cost investigation report

AllDaGearNoIdea:
Final conclusion of the #605 benchmark work. idle-cost investigation: 8h soak, fresh side-by-side vs Raycast

Our README says we idle at 3.20% CPU vs Raycast's 0.04% which is about 80x worse. From my testing, that seems to be off the mark. Full write-up in docs/explanation/idle-cost-investigation.md; raw soak data committed under benchmarks/results/idle-2026-08-07/.

Headlines:
- The published table was measured 2026-07-17 on a different machine, OS and build, with a protocol that overlaps "idle" with startup indexing and the update check (7.27% mid-index vs 0.66% actually idle on the same build).
- Repaired benchmark, current build, same machine as Raycast Beta: 0.84% vs 0.76% deep-idle CPU (a tie), ~50 wakeups/s each, we summon 2.2x faster and use 35% less memory. The one real loss is idle disk writes (487/min vs 2, from the sdk-playground dev extension fsyncing every 10s).
- 8h soak: no Rust-side leak (flat all night); the parked launcher webview creeps ~1.4MB/h, still under investigation (I'm going to do this overnight)
- Report includes a census of every timer/poll in the app with periods, the two fixes from #606 and #607 with before/after numbers, and proposals I didn't look into further (webview parking policy, extension write batching, FSEvents coalescing, the 2s lid poll).

I've not touched README.md

If/when you get time, could you please run on your machine? You seem to have a better one, and it'd make most sense to treat yours as the single source of truth.

The extended bench.sh in #605 adds machine+OS+build stamps to make clear on source for future metrics.

Feel free to not close / not approve this, but sharing for your info @Xoshbin as this was a pretty expensive agent run and it'd be good to indelibly capture / share 😓


Xoshbin:
Hey @AllDaGearNoIdea, thanks for the PRs.

I'll be taking a look at them one by one today, assuming nothing urgent comes up.

At a quick glance, the main thing that caught my eye is the 1000ms increase in cold start time. I'll test it out on my machine to see if it yields similar benchmarks, as you suggested.

PR #611, opened 2026-08-12
feat(wayland): let the compositor summon the launcher

scoop:
## Context

Wayland gives no client a global key grab, so the X11 grab behind `tauri-plugin-global-shortcut` never fires there. The launcher currently cannot be summoned at all on Wayland — which is what the README's "❌ Not supported for global hooks" reflects.

The fix isn't to make the grab work; it can't. It's to hand the key to the compositor and give it a way to drive an already-running instance:

```
bind = ALT, SPACE, exec, asyar    # Hyprland; same idea on sway, KDE, GNOME
```

## What's here

Two halves that are useless apart, which is why they're one PR.

**1. Toggle on re-exec.** The `tauri-plugin-single-instance` callback was an empty closure. It now toggles the launcher, so a second launch of the binary reveals or hides the running instance. The toggle body is lifted out of `handle_shortcut` into `toggle_launcher` — a pure move, no behaviour change on the hotkey path.

Instances carrying an `asyar://` URL are skipped: those exist to deliver a deep link, whose handler already decides whether to reveal the window (view commands do, background ones don't). Without the guard, a deep link arriving while the launcher was open would hide it.

**2. Blur grace window.** `set_focus()` cannot focus a window on Wayland — it sends an xdg-activation request and returns `Ok(())` whether or not the compositor honours it. Between `show()` and the compositor granting focus, the window is mapped but unfocused, so the `Focused(false)` handler hides the launcher the instant it's summoned. Under `follow_mouse` the compositor may never grant focus at all, because the pointer is still over whatever the user was looking at.

A blur within 250ms of a reveal is therefore treated as compositor noise. Anything later still hides, so click-away is unchanged.

Without half 2, half 1 produces a launcher that maps and hides itself in the same frame. Without half 1, half 2 fixes a race nobody can reach.

## Scope

Linux-only (`#[cfg(target_os = "linux")]`), so Windows behaviour is untouched. X11 shares the path harmlessly — its grab focuses synchronously, and no human reaches a deliberate click-away inside 250ms.

## Testing

Four unit tests on the grace-window predicate. Verified on Hyprland 0.56.2 / Arch:

- four consecutive toggles alternate cleanly
- the launcher takes keyboard focus when shown (`hyprctl activewindow` → `asyar`)
- the previously focused window regains focus when hidden
- moving focus away still hides it, so click-away survives

Before the grace window, repeated toggles left the launcher permanently hidden.

Cost is ~250ms per press: process spawn plus dynamic linking, against ~12ms for the in-process grab. Unavoidable with a re-exec, and it only applies to compositor-bound summons.

> [!NOTE]
> **Reviewing on Wayland:** the app cannot start on this branch alone because of the snap-guides abort in #610. Apply that first to exercise this. The two don't otherwise depend on each other and can merge in either order.

## Caveats

This does not make Asyar fully work on Wayland. Snippets, selection capture, window management and frontmost-app events remain broken — some by Wayland's security model. What this buys is that the app can be summoned, which is the difference between unusable and usable-with-known-gaps.

I've only tested Hyprland. The mechanisms are compositor-agnostic, but the 250ms constant is tuned against one compositor's timing and may want revisiting.

### DIFF
--- a/asyar-launcher/src-tauri/src/commands/app.rs
+++ b/asyar-launcher/src-tauri/src/commands/app.rs
@@ -67,6 +67,8 @@ pub fn prepare_show(
             y: OFFSCREEN_COORD,
         }));
         let _ = window.show();
+        #[cfg(target_os = "linux")]
+        crate::mark_launcher_shown(&state);
     }
     let _ = state;
     Ok(())
@@ -100,6 +102,8 @@ pub fn commit_show(
             .ok_or_else(|| AppError::NotFound("launcher window".to_string()))?;
         crate::launcher_placement::service::apply(&app_handle)?;
         let _ = window.set_focus();
+        #[cfg(target_os = "linux")]
+        crate::mark_launcher_shown(&state);
     }
     // Flip the visibility flag only after the panel is actually composited at
     // its final position and alpha — otherwise a concurrent showWindow() that
@@ -135,6 +139,8 @@ pub fn show(app_handle: AppHandle, state: tauri::State<'_, AppState>) -> Result<
         crate::launcher_placement::service::apply(&app_handle)?;
         let _ = window.show();
         let _ = window.set_focus();
+        #[cfg(target_os = "linux")]
+        crate::mark_launcher_shown(&state);
     }
     Ok(())
 }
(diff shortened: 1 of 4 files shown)


Xoshbin:
Thanks @scoop! Great architectural approach to solving global hotkey summoning on Wayland.

The 250ms focus grace window cleanely fixes the race condition with Wayland focus events, and re-using `tauri-plugin-single-instance` for toggling is very clean. Tested logic looks solid and unit tests pass.

PR #626, opened 2026-08-16
fix(applications): resolve and propagate localized macOS app names

kai-osthoff:
## What

Follow-up to #590. That PR made macOS apps searchable under their localized name, but on a German system Photos.app is still indexed and displayed as "Photos", never "Fotos". Two independent defects stand in the way — the second one means fixing the first alone would change nothing for existing users.

---

## 1. The name lookup cannot return a localized name

`NSFileManager displayNameAtPath:` resolves against the **calling process's** effective localization, not the user's language preference. Asyar ships English only, so macOS answers in English regardless of system language. Measured on a `de-DE` machine:

```
FileManager.displayName(atPath: "/System/Applications/Photos.app")  = "Photos.app"
Locale.preferredLanguages                                           = ["de-DE"]
Bundle.main.preferredLocalizations                                  = ["en"]   ← the cause
Bundle(path:).localizedInfoDictionary["CFBundleDisplayName"]         = "Photos"
```

Stripping `.app` then yields `"Photos"` — identical to the file stem we already had. `NSBundle localizedInfoDictionary` shares the flaw, so it is not an alternative. Both APIs look correct in isolation, which is why this slipped through.

**Fix:** read the name from where Apple actually stores it — inside the bundle:

- `Contents/Resources/InfoPlist.loctable`, one plist per language code
- `<lang>.lproj/InfoPlist.strings` for bundles predating the loctable

resolved against `sys_locale::get_locale()` — the user's preference — so the result no longer depends on how the host process is localized. Returns `None` when a bundle ships no translation for the locale, and callers keep the file stem exactly as before.

Three details the real tables force:

- **`de-DE` before `de`.** macOS reports the regional form; most bundles key on the base language, some (pt-BR) on both.
- **`CFBundleName` as fallback.** System Settings.app ships no German `CFBundleDisplayName`, only `CFBundleName`.
- **Soft hyphen stripped.** Apple embeds `U+00AD` in some names (`System\u{ad}einstellungen`). It is invisible, so a user never types it — leaving it in denies the query an exact-title match.

The lookup moves to `platform/macos/display_name.rs`, next to the other OS-level wrappers, with a `None` stub for other platforms. Linux `.desktop` files carry `Name[<lang>]=` entries and could use the same treatment later.

## 2. A corrected name never reaches an existing index

`sync_application_index` is a pure add/remove diff keyed on the app id, and the id is built from the bundle **file** name — which no change to the displayed name touches. An app already in the index appears in both the indexed and the scanned set, so it is neither added nor removed, and its stored `name` is never written again.

Verified against a real index built by a released build:

```
name    | id
Photos  | app_Photos__System_Applications_Photos.app
```

So without this second commit, part 1 changes nothing for anyone who has already run Asyar — only a fresh install or a newly installed app would pick the translation up. The same blind spot hits a user who switches macOS to another language after the first scan.

**Fix:** `refresh_display_names` reconciles the names of entries that stay. It copies the name and nothing else — `usage_count` and `last_used_at` live on the indexed item and are zero on a freshly scanned one, so swapping the whole struct (the obvious shortcut) would wipe every renamed app's frecency on the next scan. A test pins that. The write lock is now taken on every sync, since a rename-only sync must still persist.

---

## Tests

The previous test asserted only locale-independent properties, which the broken implementation satisfied — it could not have caught this. The new unit tests **inject the locale** rather than reading it from the host, so the German case is asserted on every machine and in CI:

- translated name is read from the loctable (`de` → Fotos, `en` → Photos)
- regional table wins over the base language (pt-BR)
- regional locale falls back to the base language (`de-DE`/`de_DE` → `de`)
- untranslated locale, translation-less bundle, and missing path all yield `None`
- soft hyphen stripped; a name of only invisible marks counts as absent
- a stale indexed name is refreshed; frecency survives it; unchanged and unscanned entries are left alone

Verified end-to-end on a `de-DE` system through the real code path:

```
/System/Applications/Photos.app          -> Some("Fotos")
/System/Applications/Notes.app           -> Some("Notizen")
/System/Applications/System Settings.app -> Some("Systemeinstellungen")
/System/Applications/Calendar.app        -> Some("Kalender")
/Applications/Safari.app                 -> Some("Safari")
```

Both spellings stay searchable via the existing `search_names()` plumbing from #590, which is unchanged.

`cargo fmt --check` and `cargo clippy` clean. Full suite passes apart from `file_index::deep::mdfind::tests::probe_succeeds_on_a_real_mac`, which fails identically on unmodified `main` — pre-existing and unrelated.

### DIFF
--- a/asyar-launcher/src-tauri/Cargo.lock
+++ b/asyar-launcher/src-tauri/Cargo.lock
@@ -265,6 +265,7 @@ dependencies = [
  "sha2",
  "specta",
  "specta-typescript",
+ "sys-locale",
  "sysinfo",
  "tar",
  "tauri",
--- a/asyar-launcher/src-tauri/Cargo.toml
+++ b/asyar-launcher/src-tauri/Cargo.toml
@@ -43,6 +43,9 @@ r2d2 = "0.8"
 # track newer rusqlite majors and will not link against the pin above.
 r2d2_sqlite = "0.24"
 plist = "1.7"
+# The user's language preference, which is what bundle display names must be
+# resolved against — Cocoa's own lookups answer in the host process's language.
+sys-locale = "0.3"
 tauri-plugin-notification = "2"
 tokio = { version = "1.0", features = ["full"] }
 tauri-plugin-clipboard-x = "2"
--- a/asyar-launcher/src-tauri/src/application/service.rs
+++ b/asyar-launcher/src-tauri/src/application/service.rs
@@ -567,39 +567,13 @@ pub(crate) fn bundle_file_name(path: &Path) -> String {
         .to_string()
 }
 
-/// The name macOS presents for a bundle, which is localized — `Photos.app`
-/// reads as "Fotos" on a German system. Returns `None` when the OS has
-/// nothing to add, so callers fall back to the file name.
-#[cfg(target_os = "macos")]
-pub(crate) fn localized_display_name(path: &Path) -> Option<String> {
-    use objc2_foundation::{NSFileManager, NSString};
-
-    let path_str = path.to_str()?;
-    let ns_path = NSString::from_str(path_str);
-    let display = unsafe { NSFileManager::defaultManager().displayNameAtPath(&ns_path) };
-    // Finder hides the extension by default but honours the user's
-    // "show all filename extensions" setting, so it may come back either way.
-    let display = display.to_string();
-    let trimmed = display.strip_suffix(".app").unwrap_or(&display).trim();
-    if trimmed.is_empty() {
-        None
-    } else {
-        Some(trimmed.to_string())
-    }
-}
-
-/// What the user sees for this app. Localized on macOS, the plain file name
-/// everywhere else. Deliberately separate from [`bundle_file_name`] — search
-/// matches both, but only the file name may reach `build_app_id`.
+/// What the user sees for this app: the name translated into the user's
+/// language where the bundle ships one, the plain file name otherwise.
+///
+/// Deliberately separate from [`bundle_file_name`] — search matches both, but
+/// only the file name may reach `build_app_id`.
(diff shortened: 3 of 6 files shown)


Xoshbin:
Hi @kai-osthoff,

Thank you so much for this PR and for digging into the root cause! Replacing Cocoa’s host-process-dependent `displayNameAtPath:` with direct bundle inspection is a great catch and an elegant solution. 

I really appreciate the attention to detail here—especially handling the invisible soft hyphens (`\u{ad}`) that Apple embeds in German system names like *"Systemeinstellungen"*, and ensuring `refresh_display_names` updates existing records without wiping user frecency history (`usage_count` / `last_used_at`). The test coverage is also fantastic!

While reviewing and testing against macOS bundle tables, I noticed one small edge case regarding regional locale keys in Apple's `InfoPlist.loctable` that might be worth considering:

- In stock macOS bundles (like `Photos.app` or `Siri.app`), Apple often formats regional keys with underscores (e.g., `zh_CN`, `zh_TW`, `zh_HK`, `pt_PT`, `en_GB`, `es_419`).
- For Chinese specifically, Apple’s `.loctable` only provides `zh_CN`, `zh_HK`, and `zh_TW` without a fallback bare `zh` entry.
- Because `language_candidates` currently normalizes underscores to hyphens (`zh-CN`, `zh`), lookups against tables with `zh_CN` might return `None` and fall back to the English bundle name.

Testing both hyphen and underscore variants in `language_candidates` (e.g., `["zh_CN", "zh-CN", "zh"]` or trying both `lang` and `lang.replace('-', '_')` / `lang.replace('_', '-')`) would make sure regional locales and script tags (`zh-Hans`) are fully caught as well.

I'm doing a quick smoke test by switching my macOS locale to German now. Thank you again for your fantastic work on this!

PR #632, opened 2026-08-17
fix(clipboard): give each image row its own file instead of a dead cache

kai-osthoff:
> **Stacked on #631.** Both touch `clipboardHistoryService`, so this branch is built on top of that one and its commit shows up here too — review/merge #631 first, then this diff reduces to its own single commit (`c88238b`). GitHub would not accept the stacked base directly because #631 is a cross-fork PR.

## Problem

The `clipboard_cache` mechanism for `image` entries has never worked. Confirmed at runtime in a dev build:

```
Failed to copy image to cache, using temp path: forbidden path:
…/Application Support/org.asyar.devclipboard_cache, maybe it is not allowed
on the scope for `allow-mkdir` permission in your capability file
```

Three stacked defects:

1. **Missing path separator.** `getCacheDirPath()` did `` `${appData}${CLIPBOARD_CACHE_DIR}` `` — `appDataDir()` returns no trailing slash, so the path was `org.asyar.dev` + `clipboard_cache` glued together. Invisible in tests because the `appDataDir` mock returned `'/mock/app/data/'` *with* a trailing slash the real API does not.
2. **`fs:allow-mkdir` scope** covered only the extensions directories, never `$APPDATA/clipboard_cache`.
3. **`fs:allow-copy-file` and `fs:allow-remove` were never granted at all** — so the copy failed, and every later `deleteImageFromCache` was a silent no-op.

Consequence: every `image` row pointed at the plugin's own `$APPDATA/tauri-plugin-clipboard-x/images/<hash>.png`. Previews *did* work — that path is readable under `fs:default` — but:

- **Nothing was ever deleted.** 122 MB of orphaned PNGs in the production profile on this machine; `clipboard_cache/` did not exist in either profile.
- **The plugin's path is keyed by a hash of the pixel bytes**, so two rows holding the same image share one file. Naively fixing the delete permission would have made deleting one row break the other.

## Change

Rust **moves** the plugin's freshly written PNG to `$APPDATA/clipboard_cache/<item-id>.png`, so each row owns its file outright.

A move rather than a copy: both directories live under `$APPDATA`, so the rename is O(1) and leaves **one** file on disk instead of two. The plugin re-encodes on the next identical copy anyway — it only skips writing while its hash path still exists ([`commands.rs:325`](https://github.com/cakioe/tauri-plugin-clipboard-x), `if !path.exists()`).

Rust rather than widening the webview's fs capability, matching the choice already made for thumbnails and text previews. Granting copy/rename/remove would hand the webview a general write primitive in the app data directory for the sake of one internal file move. The item id is validated Rust-side rather than trusted, so a crafted id cannot escape the cache directory.

Backward compatible by construction: `forget_image` ignores any path outside `clipboard_cache/`, so a row captured before this can never unlink a file another row still points at. And if the move fails, the row records the plugin's original path — still readable, so the preview keeps working exactly as before.

## Performance

The concern was that this sits on the clipboard capture path, which runs on every copy. Measured on this machine:

| Operation | Cost |
|---|---|
| `rename` within `$APPDATA` | O(1), microseconds |
| (for comparison) full copy of the largest cached clipboard image, 4.0 MB | ~7 ms |
| (already spent per copied image, by the plugin) decode + PNG encode | ~100 ms |

The move is free. Even the copy fallback would have been under 10% of what the plugin already spends on the same image.

## Testing

- **Rust**: 10 unit tests over `clipboard_cache` — path derivation, id validation (rejects `../../etc/passwd`, `a/b`, empty, spaces), move-not-copy, directory creation, idempotence, missing source, delete, delete-already-gone, and **refusal to delete outside the cache directory**.
- **TS**: new `clipboardCacheCommands.test.ts`; the `image cache persistence` suite rewritten against the real IPC surface instead of a `plugin-fs` mock that never matched reality.
- Full frontend suite: 3231 launcher + 660 SDK passed, after rebasing onto current main. `cargo clippy --all-targets` clean.

### Verified in a running dev build

Copying an image:
- creates `clipboard_cache/2404e4aa-6f13-40e5-9ed0-309b91d6a375.png` (named by the row's uuid),
- leaves the plugin's directory at the **same file count** — moved, not duplicated,
- and the detail pane renders the preview from the new path (`Image: 1200×1600`).

## Not included

The existing orphaned PNGs (122 MB in the production profile) are left alone — a startup sweep for files no `clipboard_items` row references is worth doing, but it is a separate change from fixing the broken path.

### DIFF
--- a/asyar-launcher/src-tauri/src/clipboard_cache/commands.rs
+++ b/asyar-launcher/src-tauri/src/clipboard_cache/commands.rs
@@ -1,0 +1,39 @@
+//! Thin Tauri command layer for [`crate::clipboard_cache`]. All logic —
+//! including the id and path validation — lives in the parent module so it
+//! is unit testable without a running Tauri app; this only resolves the app
+//! data directory and delegates.
+
+use tauri::{AppHandle, Manager};
+
+fn app_data_dir<R: tauri::Runtime>(app: &AppHandle<R>) -> Result<std::path::PathBuf, String> {
+    app.path()
+        .app_data_dir()
+        .map_err(|e| format!("app data dir unavailable: {e}"))
+}
+
+/// Moves the clipboard plugin's freshly written PNG into the history item's
+/// own cache slot and returns the new absolute path, which the row stores as
+/// its content.
+#[tauri::command]
+pub async fn clipboard_adopt_image<R: tauri::Runtime>(
+    id: String,
+    source_path: String,
+    app: AppHandle<R>,
+) -> Result<String, String> {
+    let root = app_data_dir(&app)?;
+    let dest = crate::clipboard_cache::adopt_image(&root, &id, std::path::Path::new(&source_path))?;
+    Ok(dest.to_string_lossy().into_owned())
+}
+
+/// Deletes a cached image when its history row goes away. Paths outside
+/// `clipboard_cache/` are ignored rather than rejected — legacy rows still
+/// point into the plugin's shared directory, and deleting there would break
+/// other rows holding the same image.
+#[tauri::command]
+pub async fn clipboard_forget_image<R: tauri::Runtime>(
+    path: String,
+    app: AppHandle<R>,
+) -> Result<(), String> {
+    let root = app_data_dir(&app)?;
+    crate::clipboard_cache::forget_image(&root, std::path::Path::new(&path))
+}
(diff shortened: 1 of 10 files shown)


Xoshbin:
Hey @kai-osthoff both of the PR merged. thanks a lot.

PR #633, opened 2026-08-18
fix(applications): match localized names keyed by region and script

kai-osthoff:
Follow-up to #626, from your review comment there: Apple keys its localization tables with underscores, and `language_candidates()` normalized them to hyphens — so a `zh_CN`-keyed table never matched and the lookup fell back to the English name.

You were right, and the check turned out worse than "might return `None`": for Chinese it *always* fails today.

## What the real tables contain

Aggregated over every `InfoPlist.loctable` under `/System/Applications`, `/System/Applications/Utilities`, `/System/Library/CoreServices` and `/Applications` on macOS 26.5.1 — 166 tables:

| key | tables |
|---|---|
| `en_GB`, `en_AU`, `fr_CA` | 165 |
| `zh_CN`, `zh_HK`, `zh_TW`, `pt_PT`, `es_419` | 164 |
| `pt_BR` | 136 |
| `es_US` | 60 |
| `yue_CN` | 2 |
| **bare `zh`** | **0** |

Underscore only — not one hyphen, not one script subtag, anywhere. `.lproj` directories follow the same convention (`zh_CN.lproj`, `es_419.lproj`).

## Why a second spelling isn't enough

`sys_locale` returns the preferred language as a BCP-47 tag, and for Chinese that is `zh-Hans-CN`. Plain truncation walks `zh-Hans-CN` → `zh-Hans` → `zh` — and since no bundle ships `zh_Hans` *or* a bare `zh`, every step misses however it is spelled. The script subtag has to be dropped from the middle rather than truncated off the end.

So `language_candidates()` now parses the tag into subtags instead of doing string surgery, and emits, most specific first:

```
language-script-region → language-region → language-script → implied region → language
```

each in both spellings, hyphen first, deduped. `implied_region()` covers the one language where it matters: `zh`+`Hant` → `TW`, `zh`+`Hans`/none → `CN`. It runs *before* the bare language deliberately, so a Traditional preference is never served the Simplified entry.

One parsing detail: a one-character subtag opens an extension sequence (`de-DE-u-ca-chinese`), so parsing stops there — otherwise `ca` reads as Canada.

## Validated against Apple's own resolver

Rather than trust the ordering, I asked CoreFoundation. `CFBundleCopyLocalizationsForPreferences` with Photos' real key list agrees with this chain on 19 of 22 locales — `zh-Hans-CN`→`zh_CN`, `zh-Hant-HK`→`zh_HK`, `zh-Hant`→`zh_TW`, `zh`→`zh_CN`, `de-DE`→`de`, `pt-BR`→`pt`, `en-GB`→`en_GB`, `es-419`→`es_419`, `fr-CA`→`fr_CA`, `en-US`→`en`, `he-IL`→`he`, …

The three misses are alias and likely-subtag data: `nb-NO`→`no`, `yue-Hans-CN`→`zh_CN`, `es-MX`→`es_419`. Replicating those needs CLDR tables I don't think belong in this crate, so they are documented as a known gap in the module doc. A miss costs only the fallback that has always applied — the on-disk file stem — so those users are no worse off than before this PR.

If you'd rather close that gap, the alternative is calling `CFBundleCopyLocalizationsForPreferences` directly instead of building the chain by hand. It would be exactly right, aliases included, and it does *not* have the calling-process localization flaw that made the Cocoa APIs unusable in #626. I skipped it because it costs CF FFI and `unsafe`, but happy to switch if you prefer it.

Also folded in: the loctable is now parsed once per bundle instead of once per candidate. Candidate order and the loctable-before-`.lproj` precedence per candidate are unchanged.

## Tests

Written failing first. Before the change: `zh-Hans-CN` → `None`, `en-GB` → `"Photos"` instead of `"Photos GB"`.

- `resolves_chinese_against_apples_underscore_only_keys` — a table keyed only `zh_CN`/`zh_HK`/`zh_TW`, exercised through 8 locale forms (`zh-Hans-CN`, `zh_CN`, `zh-CN`, `zh-Hant-HK`, `zh-Hans`, `zh-Hant`, `zh`, …)
- `resolves_underscore_keyed_regions_and_keeps_the_base_language_for_others` — `en_GB`/`pt_PT`/`es_419`, plus the counter-check that `en-US`, `pt-BR`, `es-ES` still land on the base language
- `language_candidates_cover_both_separators_and_script_tags` — exact ordering, dedup, extension and variant subtags
- `resolves_a_real_stock_bundle_in_an_injected_locale` — against the real `/System/Applications/Photos.app`, existence-guarded; pins no Apple translation, only that Chinese no longer resolves to the English name

Every test injects the locale rather than reading it from the host, so they are deterministic on any machine and in CI.

`cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` and `cargo doc --no-deps --document-private-items` clean. `cargo test --lib`: 3373 passed, 1 failed — `file_index::deep::mdfind::tests::probe_succeeds_on_a_real_mac`, which fails identically on an unmodified checkout of `main` here; pre-existing and environmental.

## What is not verified

This machine is `de-DE`. There was **no manual test under a Chinese, British or Portuguese system locale** — the fix is evidenced by the injected unit tests, the key inventory of the 166 system bundles, and the CoreFoundation comparison, not by an actual language switch. Worth a smoke test on your side if you have a `zh` install handy.

### DIFF
--- a/asyar-launcher/src-tauri/src/platform/macos/display_name.rs
+++ b/asyar-launcher/src-tauri/src/platform/macos/display_name.rs
@@ -24,28 +24,112 @@ pub fn localized_bundle_name(path: &Path) -> Option<String> {
 
 fn localized_bundle_name_in_locale(path: &Path, locale: &str) -> Option<String> {
     let resources = path.join("Contents/Resources");
+    let loctable = plist::Value::from_file(resources.join("InfoPlist.loctable")).ok();
+    let loctable = loctable.as_ref().and_then(plist::Value::as_dictionary);
     language_candidates(locale).into_iter().find_map(|lang| {
-        name_from_loctable(&resources.join("InfoPlist.loctable"), &lang).or_else(|| {
+        name_from_loctable(loctable, &lang).or_else(|| {
             name_from_info_plist_strings(&resources.join(format!("{lang}.lproj/InfoPlist.strings")))
         })
     })
 }
 
-/// Language keys to try, most specific first — `de-DE` before `de`. Bundles key
-/// their tables either way, and macOS itself falls back along the same chain.
+/// Language keys to try, most specific first — `de-DE` before `de`.
+///
+/// Two shapes of the same locale have to be tried, because the tag macOS
+/// reports and the key a bundle uses are written differently. `sys_locale`
+/// returns the user's preferred language as a BCP-47 tag with hyphens
+/// (`de-DE`, `zh-Hans-CN`), while Apple keys its own tables with underscores
+/// and no script subtag — every one of the 166 system `InfoPlist.loctable`s on
+/// macOS 26 uses `zh_CN`, `zh_HK`, `zh_TW`, `pt_PT`, `pt_BR`, `en_GB`,
+/// `es_419` and never a hyphen. Third-party bundles do use the hyphen form, so
+/// both are emitted, hyphen first.
+///
+/// A script subtag also has to be dropped on the way down, not just truncated
+/// off the end: Chinese arrives as `zh-Hans-CN`, and plain RFC 4647 truncation
+/// (`zh-Hans-CN` → `zh-Hans` → `zh`) never reaches `zh_CN` — and Apple ships no
+/// bare `zh` entry at all, so that chain ends in the English name. The order
+/// below matches what CoreFoundation's own resolver
+/// (`CFBundleCopyLocalizationsForPreferences`) picks when asked with Photos'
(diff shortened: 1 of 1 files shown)


Xoshbin:
Thanks for the thorough investigation and clean fix, @kai-osthoff!

The subtag parsing logic and the single-pass `.loctable` optimization are great improvements, and the injected unit tests make the resolution behavior rock-solid. Keeping this in pure Rust without pulling in CoreFoundation FFI is definitely the right trade-off here.

Merging now!

PR #636, opened 2026-08-18
fix(scripts): run the selected script on Enter in the library view

kai-osthoff:
## The bug

The Script Library view shows `Run Command  ⏎` in the bottom bar, but pressing Enter does nothing. The script only runs via ⌘K → *Run Script*.

Two separate defects:

**1. Enter never reached the view.** `ScriptLibraryView.svelte` registered a window keydown listener that only handled ArrowUp/ArrowDown. `launcherKeyboard.tryHandleViewEnter()` deliberately does *not* consume Enter here — the scripts manifest is `"searchable": false`, so it returns `false` and lets Enter bubble to a window-level listener, which is the contract built-in features rely on (clipboard-history, snippets). Scripts never registered one, so `ScriptsExtension.onViewSubmit()` — which already calls the run path — was unreachable.

**2. The hint was stale.** `PrimaryActionDisplay` falls back to the *selected search result*'s type when `activeViewPrimaryActionLabel` is null. The row that opened the view is of type `command`, hence the misleading "Run Command". Scripts never published a label, unlike clipboard-history (`Paste`), help (`Open Guide`), store (`Show Details`).

## The fix

- Extract `runSelectedScript()` from the private method on `ScriptsExtension` into `runSelected.ts`. `index.ts` imports the view, so the view cannot import back — the shared module breaks the cycle. `onViewSubmit` and the `scripts:run` action now call it too, so all three entry points share one path (including the arguments prompt).
- Handle Enter in the view's existing window listener: run the selected script, or repair a `makeExecutable` issue. Enter is consumed **only** when the selection has something to do — otherwise it keeps bubbling to the launcher.
- Publish an accurate primary action label (`Run Script` / `Make Executable` / none) and clear it on destroy, guarded so a replacement navigation cannot wipe the *incoming* view's label.

### Argument mode

`runSelectedScript()` promotes a script that declares arguments into command-argument mode, which leaves this view mounted underneath the chip row. Since the listener is window+capture it would beat the chips' own Enter, re-enter argument mode and discard what the user typed. The handler therefore stands down while `commandArgumentsService.active` is set, and for any focused text field.

## Verification

Traced the full runtime route: `handleGlobalKeydown` (window capture) → `tryRouteToActiveView` returns without consuming for built-in features → the view's capture listener acts and stops propagation → the container-level `handleKeydown` never sees it. `tryHandleSearchNavigation` bails on `viewManager.activeView`, so the launcher search input does not swallow Enter first.

- `pnpm --filter asyar test:run` — 285 files, **3252 tests passing**, including 17 new `ScriptLibraryView.test.ts` cases covering run-on-Enter, issue repair, fall-through for empty/unfixable selections, modifier keys, modal and action-popup guards, argument mode, focused text fields, and every label state.
- `pnpm format:check` and `pnpm check:design` clean.
- `svelte-check` reports only pre-existing errors; none in `built-in-features/scripts/`.

### DIFF
--- a/asyar-launcher/src/built-in-features/scripts/ScriptLibraryView.svelte
+++ b/asyar-launcher/src/built-in-features/scripts/ScriptLibraryView.svelte
@@ -3,6 +3,7 @@
   import { scriptsManager } from './scriptsManager.svelte';
   import { runSelectedScript } from './runSelected';
   import { viewManager } from '../../services/extension/viewManager.svelte';
+  import { commandArgumentsService } from '../../services/search/commandArguments';
   import {
     Badge,
     Card,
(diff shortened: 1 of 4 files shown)


Xoshbin:
Looks great, thanks for the contribution! Merged.

PR #654, opened 2026-08-21
fix(calculator): read and render numbers in the user's notation

kai-osthoff:
## The bug

Typing `61,78*1,19` into the search bar returns **735,182**.

fend — like every expression parser — only understands the `1,234.56`
convention, so it read both commas as thousands grouping and evaluated
`6178 * 119`. The answer was then grouped back with commas for display,
which is what makes the wrong number look like a plausible right one:
`735,182` reads as "735.182" to exactly the user who typed the query
that produced it.

Nothing in the calculator pipeline knew what the host locale writes, so
this was silent for every comma-decimal locale — most of Europe and
Latin America.

## The fix

The pipeline keeps a single internal notation and translates at its two
ends:

- **In** — `canonicalize_input` rewrites the query before evaluation:
  `61,78*1,19` → `61.78*1.19`, `1.234,56` → `1234.56`.
- **Out** — `localize_output` rewrites the answers after evaluation:
  `73.5182` → `73,5182`, `1,234,567` → `1.234.567`.

Every handler in between — colors, bases, percent, ratios, cooking,
dates, timezones — and fend itself keeps seeing numbers written the one
way it already understands. No handler had to learn about separators.

Two conversion rules are deliberately conservative:

- Only separators sitting **between two digits** are touched, so list
  commas survive: `rgb(255, 0, 0)`, `days until dec 25, 2026`.
- Point grouping is only honored where the reading is unambiguous —
  two or more groups (`1.234.567`) or a comma decimal behind it
  (`1.234,56`). A lone `1.234` stays a decimal, because reading it as
  grouping would turn `3.14 * 2` into `314 * 2`. Dotted dates
  (`25.12.2026`) fall out of the same rule, their groups not being
  three digits wide.

## Detection

The notation comes from the host locale via the `sys-locale` crate
already in the tree, using a CLDR-derived region table. **Region beats
language**: macOS reports `en-DE` for an English UI set to the German
region, and it is the region that decides how numbers are written. The
language is only consulted for a tag that carries no region (`de`), and
Canada is split down the language line (`fr-CA` writes `1 234,56`).

Point-decimal is the default for anything unrecognized, which is the
safe direction — a point-decimal query needs no rewriting at all, so an
undetected locale behaves exactly as it does today.

Detection is right for almost everyone and wrong for the people whose
system region and typing habits disagree. A **Number format** dropdown
(Automatic / `1,234.56` / `1.234,56`) in Settings → Extensions →
Calculator lets them say so directly.

## Tests

`cargo test calculator` — 163 passing (was 129).

Beyond the unit tests for detection, canonicalization, and
localization, `comma_locale_answers_match_the_point_locale_answers`
walks one query per handler through both notations and asserts the
comma answer is the point answer re-spelled. The pipeline only holds
together as long as every handler stays out of the separator business,
so a handler that starts reading or writing separators on its own gets
caught there rather than in a bug report.

Also run: full `pnpm --filter asyar test:run` (3274 passing),
`cargo clippy --all-targets -D warnings` (clean), `pnpm check:design`
(clean).

### DIFF
--- a/asyar-launcher/src-tauri/src/calculator/locale.rs
+++ b/asyar-launcher/src-tauri/src/calculator/locale.rs
@@ -1,0 +1,312 @@
+//! Locale-aware number notation for the calculator.
+//!
+//! Roughly half the world writes `61,78` where the other half writes
+//! `61.78`. fend — like every expression parser — only understands the
+//! `1,234.56` convention, so a query typed in a comma-decimal locale is
+//! silently misread: `61,78*1,19` parses as `6178 * 119`.
+//!
+//! Two conversions fix that, and the pipeline in between stays entirely
+//! in the canonical `1,234.56` notation:
+//!
+//! - [`canonicalize_input`] rewrites the query before evaluation
+//!   (`61,78*1,19` → `61.78*1.19`, `1.234,56` → `1234.56`).
+//! - [`localize_output`] rewrites the answer after evaluation
+//!   (`73.5182` → `73,5182`, `1,234,567` → `1.234.567`).
+
+/// Which convention a number is written in.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum NumberFormat {
+    /// `1,234.56` — point decimal, comma grouping.
+    #[default]
+    Point,
+    /// `1.234,56` — comma decimal, point grouping.
+    Comma,
+}
+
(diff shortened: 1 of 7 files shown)


Xoshbin:
Thanks for the clean implementation and thorough test coverage. 

The bidirectional translation approach keeps the internal calculator pipeline elegant while properly respecting regional number formatting. Merged!

PR #658, opened 2026-08-22
fix(linux): launch desktop applications through GIO

abdelrahman-alshurafa:
## Summary

Fix Linux application launching so `.desktop` entries are launched as applications instead of being opened as files by the system's default file handler.

This issue was discovered while testing the Linux application discovery changes from #648.

## Problem

Linux applications discovered from `.desktop` files were passed to `tauri_plugin_opener::open_path()`.

On Linux, this treats the `.desktop` file as a regular file. For example, launching:

`/usr/share/applications/AmneziaVPN.desktop`

opened the desktop entry in the text editor instead of launching AmneziaVPN.

## Fix

- Route Linux `.desktop` application paths through `gio::DesktopAppInfo`.
- Launch desktop entries using GIO's native application launch mechanism.
- Keep `tauri_plugin_opener` as the fallback for non-`.desktop` Linux paths.
- Move platform-specific application launch handling into the application service.
- Preserve the existing Windows `shell:AppsFolder` behavior.
- Preserve existing behavior for macOS and ordinary Windows paths.
- Add `gio` as an explicit Linux dependency.

The implementation intentionally relies on GIO rather than manually parsing or executing the desktop entry's `Exec=` field.

## Tests

Passed:

- `cargo fmt --check`
- `cargo clippy --tests -- -D warnings`
- `linux_desktop_path_selects_gio_launch_route`
- `linux_non_desktop_path_selects_opener_route`
- `linux_desktop_path_with_spaces_is_not_tokenized`
- `nonexistent_linux_desktop_entry_returns_controlled_error`

## Manual verification

Built a Linux release binary and tested the original failing case.

Before:
- Selecting AmneziaVPN opened `AmneziaVPN.desktop` in the text editor.

After:
- Selecting AmneziaVPN launches the application successfully.

### DIFF
--- a/asyar-launcher/src-tauri/Cargo.lock
+++ b/asyar-launcher/src-tauri/Cargo.lock
@@ -211,6 +211,7 @@ dependencies = [
  "futures-util",
  "fuzzy-matcher",
  "gdk",
+ "gio",
  "globset",
  "gtk",
  "hex",
--- a/asyar-launcher/src-tauri/Cargo.toml
+++ b/asyar-launcher/src-tauri/Cargo.toml
@@ -177,6 +177,7 @@ winreg = "0.56"
 xkbcommon = "0.7"
 gtk = "0.18"
 gdk = "0.18"
+gio = "0.18"
 zbus = { version = "4", features = ["blocking"] }
 procfs = "0.16"
 x11rb = "0.13"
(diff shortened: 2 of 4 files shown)


Xoshbin:
Thanks a lot for fixing this @abdelrahman-alshurafa! 🚀

The switch to `gio::DesktopAppInfo` is the right way to handle `.desktop` entry launches natively, and moving the launch logic into the application service layer keeps our architecture clean and consistent. Great test coverage as well.

Merging now!

PR #755, opened 2026-09-15
feat(file-search): native file previews — Quick Look panel, Open action, unified OS-driven preview chain 

### LINKED ISSUE #741: [Bug]: Preview of PDF files
### Asyar Version

v0.1.1-45

### Platform / OS

macOS

### Description

The PDF files are not previewing. This might be a me problem but this is what heppening:

Image

akib1689:
Fixes #741 — supersedes #752 (consolidated with the remaining file-preview work).

## The bug (#741)

Selecting a PDF in File Search dumped raw file bytes (`%PDF-1.7 …`) into the
preview pane. Root cause: PDFs share the `Document` bucket with `.txt`/`.md`,
and the text preview read raw bytes with no binary detection.

## What this PR does

**1. Preview can never show raw bytes again**
`read_text_preview` now sniffs the first 8 KiB (`%PDF` magic, ZIP/NUL
signatures) and resolves binary content to `null`. The frontend runs one
unified priority chain with no extension lists:

    text read → real text? → text pane
    binary   → OS thumbnail (Quick Look on macOS, null elsewhere)
    nothing  → "No preview available"

**2. Native Quick Look panel (macOS)**
Replaces the old `qlmanage -p` spawn — Apple's *debug* tool whose window
titles read `[Debug] <file>`. The Files view now hosts the real shared
`QLPreviewPanel` (same panel Finder uses): proper title bar, Escape closes,
Space toggles, full-res paging. No new dependencies — objc2 runtime bindings
following the existing `platform/macos` patterns.

**3. Open with the system default app — actually works**
`Enter`/open previously used the webview-scoped opener, which silently
rejected any `$HOME` path. All open flows now go through the host-context
Rust opener (`opener_open_path`): PDF → Preview.app, `.csv` → Numbers,
etc. The Files view claims "Open" as its bottom-bar primary action
(same `viewManager` pipeline the Scripts view uses), which also fixes the
stale "Run" hint leaking in from root search.

**4. Housekeeping**
- `.csv` classified as `Document` (was falling into `Other`)
- Preview state fully cleared on selection change (stale text could hide a
  newly fetched thumbnail)
- Quick Look ⌘K row shows the `␣` (Space) shortcut hint

## Testing

- Rust: binary-sniff unit tests (`%PDF` magic, NUL/ZIP, sniff-window edges),
  preview command validation, FileType classification — `cargo test`,
  clippy `-D warnings`, fmt
- Frontend: label lifecycle tests, IPC wrapper tests — 35/35 in file-search,
  full workspace suite, prettier, design-system check
- Manual QA on macOS (Apple Silicon): PDF/.xlsx/.docx inline thumbnails,
  native panel open/toggle/Escape, Enter-opens-in-Preview.app, csv as text

## Notes

- Non-macOS platforms degrade gracefully (thumbnail generators return null →
  caption); the extension-facing `files_read_text` API is unchanged


akib1689:
@Xoshbin Can you please look into this? mac os has quick look. any linux utility that we can use to get some sort of preview image? This would be an awesome edition.

Xoshbin:
hi @akib1689 thank you for opening the pr. I'm currently looking at the pr and been for the past two hours I need to make sure it's all good before replying to you and merging it. thank you again.

Xoshbin:
Hi @akib1689,

Thank you for putting this together and for your patience while I went through the changes!

I really like this architecture — moving the binary sniffing into Rust (`read_text_bounded`) completely eliminates the raw byte dump on PDFs and Office docs, and hosting the real `QLPreviewPanel` makes macOS file previews feel truly native. I also verified extension isolation is preserved, and path validation in `read_text_preview` remains solid.

I have just a couple of small fixes and a security hardening check to request before I get this merged:

---

### 1. Missed `openerService` in ⌘K Action (`index.ts`)
In `asyar-launcher/src/built-in-features/file-search/index.ts:240`, inside `id: 'file-search:quick-look'`, the fallback is still using the direct `@tauri-apps/plugin-opener` import:

```typescript
// asyar-launcher/src/built-in-features/file-search/index.ts:236-241
execute: async () => {
  const f = getSelectedFile();
  if (!f) return;
  const ok = await quickLookPath(f.path);
  if (!ok) await openPath(f.path); // <-- Still calls @tauri-apps/plugin-opener
},
```

On non-macOS platforms, this fallback will hit Tauri's webview scope restriction and silently fail for `$HOME` paths. Could you update line 240 to:

```typescript
if (!ok) await openerService.openPath(null, f.path);
```

Once updated, `openPath` can be removed from `import { openPath, revealItemInDir } from '@tauri-apps/plugin-opener';` at the top of `index.ts`.

---

### 2. Unused Import in `DefaultView.svelte`
In `asyar-launcher/src/built-in-features/file-search/DefaultView.svelte:3`, `import { openPath } from '@tauri-apps/plugin-opener';` is no longer used since line 213 was switched to `openerService.openPath`. We can safely remove that import.

---

### 3. Security Hardening: Absolute Path Check in `quick_look_path`
In `asyar-launcher/src-tauri/src/commands/file_manager.rs:180`, `quick_look_path` currently checks:

```rust
let path = Path::new(&path_str).to_path_buf();
if !path.exists() { ... }
```

Cocoa's `[NSURL fileURLWithPath:]` requires an absolute path; passing a relative path can resolve unexpectedly against the application bundle directory. Could you add an absolute path check matching `validate_show_path`?

```rust
if !path.is_absolute() {
    return Err(AppError::Other(format!("Path must be absolute: {path_str}")));
}
```

---

### Optional Rust Polish (Nice-to-have)
- **Directory Preview Flag:** In `platform/macos/quicklook.rs:66`, passing `path.is_dir()` (or `Bool::from(path.is_dir())`) instead of hardcoded `isDirectory: false` allows macOS to show proper folder preview cards if Quick Look is triggered on a directory.
- **Non-blocking Tokio Channel:** In `quick_look_path`, switching `std::sync::mpsc` + `rx.recv()` to `tokio::sync::oneshot` + `rx.await` avoids synchronously blocking a Tokio worker thread while waiting for the main thread.

---

### Regarding your question on Linux previews
On Linux, there is no single OS-level standard for a standalone Quick Look popup panel (GNOME uses `sushi`, KDE uses its own preview framework, and behaviors differ across X11 and Wayland compositors).

However, Asyar's Linux backend already integrates with the Freedesktop thumbnail specification (`~/.cache/thumbnails/`, `tumbler`, `ffmpegthumbnailer`, `pdftoppm`). Because your new priority chain in `DefaultView.svelte` falls back to `getFileThumbnail`, Linux users will automatically get rich inline previews (images, video stills, and PDF front pages) rendered right inside the detail pane without needing an external window.

---

Once those small items are addressed, I'll be ready to merge this in. Thanks again for the great contribution!

akib1689:
@Xoshbin done. I really hope this project goes to v1 and prod build someday. It'll be a great competitor to raycast.

Xoshbin:
Thanks for the quick turnaround and for taking care of all those details, @akib1689! Everything tested clean across all platforms, and I've just merged this into `main`.

Native Quick Look and the unified preview chain are huge upgrades for File Search. 

I also really appreciate your kind words and encouragement about the project reaching v1! There is a lot of exciting work underway to get Asyar there, and contributions like this make a huge difference. Thanks again for your support and great work!

PR #758, opened 2026-09-15
fix(cli): explain common manifest type and trigger errors

Matrikzin:
Extension authors can confuse the `text` command argument type with the `textfield` preference type, or provide a space-separated keyword list in `commands[].trigger`. Validation currently lists the allowed types or says the trigger must be alphanumeric, without explaining these corrections.

Add targeted hints for both text-type mix-ups and describe triggers as a single non-empty keyword, including the allowed ASCII characters and a valid example. These diagnostics are shared by `asyar validate` and `asyar publish`. Accepted manifest values and publishing behavior are unchanged.

Update the publishing guide and preferences reference with the same corrections. Add regression coverage for extension- and command-level preferences, arguments, invalid triggers, and valid inputs.

## Example

For `"trigger": "cep endereco"`, the diagnostic now reads:

```text
commands[0].trigger: must be a single non-empty keyword using only A-Z, a-z, 0-9, "_" or "-" (e.g., "search-cep"); spaces and keyword lists are not allowed
```

For a preference declared with `"type": "text"`, the existing error now includes:

```text
Use "type": "textfield" for a text preference; "text" is a command argument type.
```

## Validation

Verified on Windows against upstream commit `3942af2a5ba2ae0080c9d763b1557fb9e6a02dae`:

- Test-first verification: nine diagnostic assertions failed before implementation; all 131 manifest tests passed after implementation (16 added cases).
- SDK TypeScript and CLI build passed.
- Compiled CLI smoke checks passed: invalid fixture exits with code 1 and corrective messages in both `validate` and `publish`, before build/authentication; corrected fixture passes `validate` with code 0.
- Standard Prettier check passed for all four changed files.
- Design compliance passed (191 files).
- Full SDK suite: 713 passed, 3 failed in the existing `cli/commands/link.test.ts` unlink tests. They mock the manifest module and compare hard-coded POSIX paths against Windows `path.join` output; the changed validator is not executed by those tests.
- Full launcher suite (run separately from the workspace root): 3480 passed, 4 failed. Two `external-bin-provisioning` assertions assume LF in the unchanged release workflow; two `feedbackBoundary` assertions compare Windows relative paths with POSIX allowlist entries. These failures are outside the changed files.
- `pnpm check:ci` was attempted and stopped at the workspace format check because this Windows checkout uses CRLF. A read-only full-workspace Prettier check with `--end-of-line auto` passed.
- No Rust source or generated bindings changed.


Xoshbin:
Thanks for the contribution, @Matrikzin! 

Clear, actionable diagnostics make a huge difference for extension developers getting started with the SDK, and I really appreciate the thorough test coverage and documentation updates included with this.

I've merged this into `main`, and it will be included in the next SDK release!

